PR #1, opened 2016-08-17
Clang errors and examples

brianyu28:
I've added the error messages that @annieechen and I worked on last weekend, and put them into `clang.py`. I also added an example of each error message to the `examples/c/` directory for reference, with a comment atop each file explaining which error message the program is meant to generate (updated `1.c`, `2.c`, and `3.c` with comments too). Not sure if you have a better system in mind for organizing/labeling the examples, though.

Also, I noticed that the error messages already written in `clang.py` tested for `warning:` in `lines[0]` sometimes. Given that `make` calls `clang` with the `-Werror` flag, I changed occurrences of `warning` to `(?:warning|error)` so that the regex will catch cases where the `-Werror` flag is used as well.


dmalan:
Thank you!

PR #2, opened 2016-08-26
MySQL Schema, Matchers, Filenames

brianyu28:
Changes:

- configuration for MySQL tables to store `inputs` and `outputs`. @annieechen is working on the code to populate those databases.
- new matchers for some bash commands (rm confirmation, mv and cp overwrites), and additional clang/make messages.
- added filenames to `after` text whenever line numbers are mentioned.


model.py
```diff
@@ -0,0 +1,41 @@
...
+    os.environ['DB_HOST'],
+    os.environ['DB_USER'],
+    os.environ['DB_PASS'],
+    os.environ['DB_NAME']
+)
+cur = con.cursor()
+
+# disable 'table already exists' warnings
+filterwarnings('ignore', category=mdb.Warning)
+
```
dmalan: Is this necessary if you're using `IF NOT EXISTS` anyway?

model.py
```diff
@@ -0,0 +1,41 @@
...
+from warnings import filterwarnings
+
+# establish connection to database
+con = mdb.connect(
+    os.environ['DB_HOST'],
+    os.environ['DB_USER'],
+    os.environ['DB_PASS'],
+    os.environ['DB_NAME']
+)
+cur = con.cursor()
```
dmalan: Mind changing `DB_` to `MYSQL_`?

model.py
```diff
@@ -0,0 +1,41 @@
...
+
+# disable 'table already exists' warnings
+filterwarnings('ignore', category=mdb.Warning)
+
+def configure():
+    inputs_schema = '''
+    CREATE TABLE IF NOT EXISTS inputs (
+        id BIGINT NOT NULL AUTO_INCREMENT,
+        cmd TEXT,
```
dmalan: Saw your Asana. Why don't we just cut this off at 1024 chars for now, which feels like more than enough, certainly for this use case. @glennholloway

model.py
```diff
@@ -0,0 +1,41 @@
...
+        script TEXT NOT NULL,
+        username VARCHAR(32),
```
dmalan: `NULL`

model.py
dmalan: Might as well allow this to be `NULL` too.

model.py
```diff
@@ -0,0 +1,41 @@
...
+        created DATETIME,
+        PRIMARY KEY (id)
+    ) ENGINE=InnoDB;
+    '''
+    
+    outputs_schema = '''
+    CREATE TABLE IF NOT EXISTS outputs (
+        id BIGINT NOT NULL AUTO_INCREMENT,
+        input_id BIGINT,
+        output TEXT,
```
dmalan: `TEXT` feels better here, since it's even more arbitrary output.

dmalan:
Aok to merge once comments addressed!

helpers/bash.py
```diff
@@ -37,3 +37,36 @@ def help(lines):
         else:
             after.append("Did you remember to make `{}` \"executable\" with `chmod +x {}`?".format(matches.group(1), matches.group(1)))
         return (lines[0:1], after)
+    
+    # Remove file
+    matches = re.search(r"^rm: remove regular (?:empty )?file ‘(.+)’\?", lines[0])
```
dmalan: Mind ensuring that each matcher's comment matches the gist of the error message being matched, verbatim, capitalization too? E.g., let's write this as `# remove regular file, remove regular empty file` so we can search more easily.

Do keep each matcher alphabetically in the file too. Easier to skim and conclude faster whether we have a regex or not already.

helpers/bash.py
```diff
@@ -37,3 +37,36 @@ def help(lines):
...
+    if matches:
+        empty = re.search(r"^rm: remove regular empty file", lines[0])
+        after = [
+            "The command you typed will remove the file `{}`.".format(matches.group(1))
+        ]
+        if empty:
+            after.append("`{}` is an empty file.".format(matches.group(1)))
```
dmalan: Let's make this more useful information, rather than just offering up the fact. E.g., I'd append "(which is empty anyway)" to the earlier message.

helpers/bash.py
```diff
@@ -37,3 +37,36 @@ def help(lines):
         else:
             after.append("Did you remember to make `{}` \"executable\" with `chmod +x {}`?".format(matches.group(1), matches.group(1)))
         return (lines[0:1], after)
+    
+    # Remove file
+    matches = re.search(r"^rm: remove regular (?:empty )?file ‘(.+)’\?", lines[0])
+    if matches:
+        empty = re.search(r"^rm: remove regular empty file", lines[0])
+        after = [
+            "The command you typed will remove the file `{}`.".format(matches.group(1))
```
dmalan: Let's change "remove" to the clearer "delete"?

helpers/bash.py
```diff
@@ -37,3 +37,36 @@ def help(lines):
...
+        ]
+        if empty:
+            after.append("`{}` is an empty file.".format(matches.group(1)))
+        after.append("If you wish to remove `{}`, type `y` and press return.".format(matches.group(1)))
```
dmalan: Do PCs tend to have Return or Enter? Is it usually capitalized?

helpers/bash.py
```diff
@@ -37,3 +37,36 @@ def help(lines):
...
+        after.append("Typing `n` and pressing return will cancel the operation.")
+        return (lines[0:1], after)
+    
+    # mv overwrite
```
dmalan: this should go in a `mv.py` file. And comment can just be "overwrite".

helpers/bash.py
dmalan: PS, should go in `rm.py`.

helpers/bash.py
```diff
@@ -37,3 +37,36 @@ def help(lines):
...
+            "Typing `n` and pressing return will cancel the operation."
+        ]
+        return (lines[0:1], after)
+    
+    # cp overwrite
+    matches = re.search(r"^cp: overwrite ‘(.+)’\?", lines[0])
+    if matches:
+        after = [
+            "You are copying a file to `{}`, but there is already a file with the same name there.".format(matches.group(1)),
+            "To continue with the copy operation and replace the old file, type `y` and press return.",
```
dmalan: Let's avoid heavy language like "copy operation" and make more TF-like: "To copy {}, replacing the old version, type..."

helpers/clang.py
```diff
@@ -125,3 +134,23 @@ def help(lines):
     if matches:
         after = ["It seems that the variable `{}` is never in your program. Try either removing it altogether or using it.".format(matches.group(1))]
         return (lines[0:1], after)
+    
+    # more % than arguments
```
dmalan: Do write comment in a way that matches regex more closely.

helpers/bash.py
brianyu28: Did some searching, "enter" is more common on PCs, though not a whole lot of consistency across keyboards with regards to capitalization (some omit the word "enter" altogether in favor of just an arrow symbol).

curiouskiwi:
Might want a "shadow declaration" help?  If, for example, the student uses a comma rather than a semicolon in their for loop:  `for (int i = 0, i < d; i++)`, clang will spit out: `error: declaration shadows a local variable [-Werror,-Wshadow]`

model.py
brianyu28: Despite `IF NOT EXISTS`, the `MySQLdb` Python library still prints out a warning that a table already exists when trying to create it, so I included this line to suppress the warning output.

brianyu28:
@dmalan : finished the fixes per your comments from earlier today, this now includes @annieechen's code for inserting into the `inputs` and `outputs` tables, if you'd like to take a look at that before merging.

model.py
```diff
@@ -38,4 +38,13 @@ def configure():
     
     cur.execute(inputs_schema)
     cur.execute(outputs_schema)
+
+def insert_input(inputstring):
+    # will need to sanitize
+    statement = (
+        "INSERT INTO inputs (cmd, script)"
+        "VALUES (%s, %s)"
```
dmalan: Does this not need a space after `script)` if these strings get concatenated?

model.py
```diff
@@ -38,4 +38,13 @@ def configure():
     
     cur.execute(inputs_schema)
     cur.execute(outputs_schema)
+
+def insert_input(inputstring):
+    # will need to sanitize
```
dmalan: Better to call `inputstring` `cmd` if that's what it is.

model.py
```diff
@@ -38,4 +38,13 @@ def configure():
...
+    statement = (
+        "INSERT INTO inputs (cmd, script)"
+        "VALUES (%s, %s)"
+    )
+    data = (inputstring, "clang")
```
dmalan: Why is `clang` hardcoded here? `script` should be the `stdout/stderr` that got POSTed?

model.py
dmalan: Looks like you already are escaping? http://stackoverflow.com/a/7929438/5156190

model.py
```diff
@@ -39,12 +39,14 @@ def configure():
     cur.execute(inputs_schema)
     cur.execute(outputs_schema)
 
-def insert_input(inputstring):
+def insert_input(cmdtype, inputstring):
```
dmalan: Better to call these `cmd` and `script` for consistency.

dmalan:
@annieechen to avoid

`Merge remote-tracking branch 'upstream/master'`

in history, try `git pull --rebase` instead of just `git pull`.

dmalan:
@brianyu28 @annieechen, in future, do create separate PRs for different features. E.g., a PR for MySQL stuff, a PR for `clang` stuff, a PR for `rm` stuff, etc. Else everything ends up on one big PR, and smaller merges get held up.

application.py
```diff
@@ -43,7 +43,14 @@ def index():
 
                 # helpful response
                 if help:
-                    return render_template("helpful." + format, before="\n".join(help[0]), after="\n".join(help[1]))
+                    # send to db
+                    cmd = "clang"
```
dmalan: Why hardcoded?

application.py
```diff
@@ -43,7 +43,14 @@ def index():
...
+                    inputstring = "\n".join(help[0])
```
dmalan: `cmd`?

application.py
```diff
@@ -43,7 +43,14 @@ def index():
...
+                    outputstring = "\n".join(help[1])
```
dmalan: `script`?

model.py
```diff
@@ -39,14 +39,24 @@ def configure():
...
     # will need to sanitize
-    statement = (
+    inputstatement = (
         "INSERT INTO inputs (cmd, script)"
         "VALUES (%s, %s)"
     )
     data = (cmdtype, inputstring)
-    cur.execute(statement, data)
+    cur.execute(inputstatement, data)
     con.commit()
```
dmalan: If this is a transaction, should execute both statements first before committing. And need to check return value, no, to ensure that insert succeeds and `LAST_INSERT_ID()` will work as intended, rolling back transaction if not?

application.py
```diff
@@ -44,15 +44,20 @@ def index():
                 # helpful response
                 if help:
                     # send to db
-                    cmd = "clang"
+                    cmd = helper
                     inputstring = "\n".join(help[0])
                     outputstring = "\n".join(help[1])
                     print(type(inputstring))
```
dmalan: Why printing?

application.py
```diff
@@ -44,15 +44,20 @@ def index():
...
-                    model.insert_inputoutput(cmd, inputstring, outputstring)
+                    model.insert_input(cmd, inputstring)
+                    model.insert_output(outputstring)
                     return render_template("helpful." + format, before=inputstring, after=outputstring)
                 
 
         # unhelpful response
+        # send only inputs to DB
+        cmd = "unknown"
```
dmalan: Should be `NULL`.

model.py
annieechen: Will do

model.py
annieechen: I think this is an outdated commit, I've taken out that comment

model.py
annieechen: I think this is an outdated diff as well.

model.py
annieechen: will do

application.py
annieechen: I think this is an outdated diff as well.

application.py
annieechen: I think this is outdated as well

model.py
annieechen: Will do

application.py
annieechen: Outdated as well (was debugging)

dmalan: K, in future, do be sure not to include work in progress in PRs. Best to work on changes on new branches!

annieechen: To clarify, I should checkout a new branch, make all in progress commits on that branch, and then when ready, merge to master? (Sorry for asking, but I want to be sure I understand)

dmalan: Not quite. In general, when working on a "feature" (e.g., adding MySQL INSERTs), best to do, e.g.:

```
git checkout -b mysql-support
```

and then do all your work and commits on that branch. Then, when it's in good shape, open a PR from that branch to `cs50/help50` so that we can review only those MySQL-related commits all at once, then merge them into `master`. Meanwhile, if you're simultaneously working on some matcher, best to do, e.g.:

```
git checkout master
git checkout -b more-clang-matchers
```

and open a new PR for that too. Make sense?

Right now we've a huge number of changes all in one PR, which means we can't/shouldn't merge until absolutely everything is done.

annieechen: Makes sense. Do I have to do anything with the commits on that branch? (Squash?). Also, I'm currently working off a fork of @brianyu28's fork. I suspect this is also the wrong workflow?

dmalan: Once you open a PR from your branch to `cs50/master`, any subsequent commits to your branch will automatically appear in the PR (until it's closed and they're merged in).

Hm, not sure about forks of forks. Probably better to work off of a fork of ours, so that when you pull in updates from master, they're coming from the official master branch.

model.py
annieechen: @dmalan, does this mean there should only be one function that inserts an output if passed in and the insert succeeded, but otherwise only an input?

dmalan: Yup, one function sounds fine. And if `output` is `None`, then you can skip the `INSERT` into `outputs`!

brianyu28:
Noticing that this branch apparently has conflicts now, would it be easier for merging purposes to cancel this PR, separate the code into multiple branches, and then submit new PRs? Or would you prefer sticking with this one until all changes are ready?

dmalan:
Aok to close and resubmit as separate ones, especially if that gets some of the updates out faster!

Issue #3, opened 2016-08-27
shadow declaration clang error


curiouskiwi:
Might want a "shadow declaration" help?  If, for example, the student uses a comma rather than a semicolon in their for loop:  `for (int i = 0, i < d; i++)`, clang will spit out: `error: declaration shadows a local variable [-Werror,-Wshadow]`

curiouskiwi:
That typo would also give this error: `expected ';' in 'for' statement specifier` which has been added to help50 so probably not the best example.  Here's another example I've seen:
```

#include <cs50.h>
#include <stdio.h>

int main(void)
{
    int height = 0;
    do
    {
        int height = get_int();
        printf("you gave me %d\n", height);
    }
    while (height > 23);
}
```

Make will return:
 ```
shadow.c:9:13: error: declaration shadows a local variable [-Werror,-Wshadow]
        int height = get_int();
            ^
shadow.c:6:9: note: previous declaration is here
    int height = 0;
        ^
1 error generated.
make: *** [shadow] Error 1
```

brianyu28:
Added as of https://github.com/cs50/help50/commit/9a344a3672307a04b49a65f19af996477ee67a57

dmalan:
Thanks, @curiouskiwi, @brianyu28!

PR #4, opened 2016-08-28
add declaration shadows matcher

brianyu28:
Addressing Issue https://github.com/cs50/help50/issues/3, 

Adding a matcher to `clang.py` to address `declaration shadows a local variable` error. This one's a little bit more involved than most of the others for a few reasons:

- it checks to see whether the error message is the result of a for loop, in which case the helper message needs to be different (because the student probably just used `,` instead of `;` in their loop definition).
- if the error wasn't generated from a for loop, then we try to get the line number of the variable's previous declaration if possible, in order to provide a more useful helper.

Separated this into its own branch per your comment on workflow.

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -5,17 +5,17 @@ def help(lines):
     matches = re.search(r"^([^:]+):(\d+):\d+: (?:warning|error): declaration shadows a local variable", lines[0])
     if matches:
         after = [
-            "On line {} of `{}`, it seems that you're trying to create a new variable which has already been created.".format(matches.group(2), matches.group(1))
+            "On line {} of `{}`, it seems that you're trying to create a new variable that has already been created.".format(matches.group(2), matches.group(1))
         ]
         
         # check to see if declaration shadowing is due to for loop with commas instead of semicolons
         if len(lines) >= 2:
-            for_loop = re.search(r"^\s*for[(\s]", lines[1])
+            for_loop = re.search(r"^\s*for\s*\(", lines[1])
             if for_loop:
-                after.append("If you meant to create a for loop, be sure that each part of the for loop is separated with a semicolon (;) rather than a comma (,).")
+                after.append("If you meant to create a `for` loop, be sure that each part of the `for` loop is separated with a semicolon rather than a comma.")
                 return (lines[0:2], after)
         
-        # see if we can get the line number of the previous declaration of the varaible
+        # see if we can get the line number of the previous declaration of the variable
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -1,5 +1,37 @@
...
+    # declaration shadows a local variable
+    matches = re.search(r"^([^:]+):(\d+):\d+: (?:warning|error): declaration shadows a local variable", lines[0])
+    if matches:
+        after = [
+            "On line {} of `{}`, it seems that you're trying to create a new variable which has already been created.".format(matches.group(2), matches.group(1))
+        ]
+        
+        # check to see if declaration shadowing is due to for loop with commas instead of semicolons
+        if len(lines) >= 2:
+            for_loop = re.search(r"^\s*for[(\s]", lines[1])
```
dmalan: Hmm, sure the `[]` character class has the right semantics? Is `...for\s*\(` a closer match for what you're looking for?

helpers/clang.py
```diff
@@ -1,5 +1,37 @@
...
+            if for_loop:
+                after.append("If you meant to create a for loop, be sure that each part of the for loop is separated with a semicolon (;) rather than a comma (,).")
```
dmalan: ``for``

helpers/clang.py
```diff
@@ -1,5 +1,37 @@
...
+                return (lines[0:2], after)
+        
+        # see if we can get the line number of the previous declaration of the varaible
```
dmalan: varaible

helpers/clang.py
```diff
@@ -1,5 +1,37 @@
...
+            if prev:
+                prev_declaration_line = prev.group(2)
+                prev_declaration_file = prev.group(1)
+        
+        omit_suggestion = "If you meant to use the variable you've already declared previously"
+        if prev_declaration_line and prev_declaration_file:
+            omit_suggestion += " (on line {} of `{}`)".format(prev_declaration_line, prev_declaration_file)
+        omit_suggestion += ", try getting rid of the data type of the variable on line {} of `{}`. You only need to include the data type when you first declare a variable.".format(matches.group(2), matches.group(1))
+        after.append(omit_suggestion)
+        after.append("Otherwise, if you did mean to declare a new variable, try changing its name to a name which hasn't been used yet.")
```
dmalan: s/which/that/

brianyu28:
Made tweaks per your comments, should be addressed in the most recent commit.

PR #7, opened 2016-08-28
mv.py

brianyu28:
Matcher for `mv` overwrite

### DIFF
--- a/helpers/mv.py
+++ b/helpers/mv.py
@@ -3,11 +3,22 @@ def help(lines):
    
     # $ mv foo bar
     # mv: overwrite ‘bar’?
+    #
+    # $ mv foo baz/foo
+    # mv: overwrite ‘baz/foo’?
     matches = re.search(r"^mv: overwrite ‘(.+)’\?", lines[0])
     if matches:
-        after = [
-            "You are moving a file to `{}`, but there is already a file with the same name there.".format(matches.group(1)),
-            "To continue with the move operation and replace the old file, type `y` and press return.",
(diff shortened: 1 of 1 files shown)


brianyu28:
Looks like we get output like:

```
$ mv bar baz/bar
mv: overwrite ‘baz/bar’?
```
which could allow us to infer that such a command is a movement rather than a renaming. However, a situation like:

```
$ mv baz/foo baz/bar
mv: overwrite ‘baz/bar’?
```

would produce the same message from `mv`, but it's a renaming. Any good way to handle that?

dmalan:
Hm, how about "renaming" by default, else "moving" if the destination has a slash, even though the latter might still be a renaming? Guessing the example you proposed won't actually happen often in practice.

brianyu28:
Added the logic for the distinction in the most recent commit.

helpers/mv.py
```diff
@@ -3,11 +3,22 @@ def help(lines):
...
-            "Typing `n` and pressing return will cancel the operation."
-        ]
+        
+        # if a "/" is present in the destination path, assume moving a file; otherwise, renaming a file
+        moving = "/" in matches.group(1)
+        
+        interpretation = "You are "
+        interpretation += "moving" if moving else "renaming"
+        interpretation += " a file to `{}`, but there is already a file with the same name".format(matches.group(1))
+        interpretation += " there." if moving else " in the current directory."
```
dmalan: Minor but seems inconsistent to have a trailing white space in your first string but then leading white space later?

brianyu28:
Fixed.

PR #9, opened 2016-08-28
add mysql insertions for inputs and outputs

annieechen:
(this currently has a conflict, but I think once @brianyu28 merges his pr for setting up the database, should be fine)

adds logging of inputs and outputs. If input didn't have a helpful response, output isn't logged.


annieechen:
made the style tweaks and error checking per comments, I had some questions on proper style, making inline comments

application.py
```diff
@@ -37,9 +43,20 @@ def index():
...
+                    output = "\n".join(help[1])
+                    model.insert_io(cmd, script, output)
+                    return render_template("helpful." + format, before=script, after=output)
+                
 
         # unhelpful response
+        # send only inputs to DB
+        cmd = None 
+        output = None
+        model.insert_io(cmd, script, output)
```
annieechen: I'm not sure if here, it would be better style to instead insert w/

`model.insert_io(None, script, None)`?

annieechen:
Just realized I may have an incorrect assumption regarding PRs. Once Brian merges his earlier commits ( for instance adding new masters), I assumed only my correct commits would have been in the PR. Closing and adding a more specific PR

dmalan:
K, do merge `cs50/master` into your own branch ideally, so that when you PR, there aren't conflicts.

application.py
dmalan: Don't think you want to hardcode `cmd` as `None`? Its value should be `None` only if it's not in the POST.

dmalan: And I think you can do

`def insert_io(cmd, script, output=None)`

so that you can simply call `insert_io` with two arguments in that case?

dmalan: PS, instead of `insert_io` which doesn't really abstract away what's happening, a better method name would probably be `log`, since that's really what you're doing. That we're INSERTing into a DB is really just an implementation detail.

annieechen:
Somehow, my fork of a fork is making it very difficult to get cs50/master into my branch (it's pulling from brianyu28/master instead, and I can't figure out a way to reset it). I'm going to wait until cs50/help50 matches @brianyu28/help50, submit a PR then, and then once it's merged, delete my fork of a fork and fork the main repo again. In the meantime, will make fixes based on your comments.

dmalan:
K!

PR #12, opened 2016-08-28
MySQL configuration

brianyu28:
This PR just creates MySQL tables if they don't already exist upon program launch; it doesn't actually insert anything into the tables just yet (@annieechen is working on that, and will submit a separate PR for that after this one is merged).

### DIFF
--- a/model.py
+++ b/model.py
@@ -1,0 +1,42 @@
+import MySQLdb as mdb
+import os
+from warnings import filterwarnings
+
+# establish connection to database
+con = mdb.connect(
+    os.environ['MYSQL_HOST'],
+    os.environ['MYSQL_USER'],
+    os.environ['MYSQL_PASS'],
+    os.environ['MYSQL_DATABASE'],
+    charset='utf8'
+)
+cur = con.cursor()
+
+# disable 'table already exists' warnings
+filterwarnings('ignore', category=mdb.Warning)
+
+def configure():
+    inputs_schema = '''
+    CREATE TABLE IF NOT EXISTS inputs (
+        id BIGINT NOT NULL AUTO_INCREMENT,
+        cmd VARCHAR(1024) NULL,
+        script TEXT NOT NULL,
+        username VARCHAR(32) NULL,
+        created DATETIME DEFAULT CURRENT_TIMESTAMP,
+        PRIMARY KEY (id)
+    ) ENGINE=InnoDB;
+    '''
+    
+    outputs_schema = '''
+    CREATE TABLE IF NOT EXISTS outputs (
+        id BIGINT NOT NULL AUTO_INCREMENT,
+        input_id BIGINT,
+        output TEXT,
+        PRIMARY KEY (id),
+        FOREIGN KEY (input_id) REFERENCES inputs(id) ON DELETE CASCADE
+    ) ENGINE=InnoDB;
+    '''
+    
+    cur.execute(inputs_schema)
+    cur.execute(outputs_schema)
+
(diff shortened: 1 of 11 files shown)


model.py
```diff
@@ -0,0 +1,42 @@
+import MySQLdb as mdb
+import os
+from warnings import filterwarnings
+
+# establish connection to database
+con = mdb.connect(
+    os.environ['MYSQL_HOST'],
+    os.environ['MYSQL_USER'],
```
dmalan: Sorry, should be `_USERNAME` and `_PASSWORD` now.

model.py
```diff
@@ -0,0 +1,42 @@
...
+    os.environ['MYSQL_PASS'],
+    os.environ['MYSQL_DATABASE'],
+    charset='utf8'
+)
+cur = con.cursor()
+
+# disable 'table already exists' warnings
+filterwarnings('ignore', category=mdb.Warning)
```
dmalan: Just to recap, is this necessary, given that your SQL query already has `IF NOT EXISTS`?

model.py
```diff
@@ -0,0 +1,42 @@
...
+        input_id BIGINT,
+        output TEXT,
+        PRIMARY KEY (id),
+        FOREIGN KEY (input_id) REFERENCES inputs(id) ON DELETE CASCADE
+    ) ENGINE=InnoDB;
+    '''
+    
+    cur.execute(inputs_schema)
+    cur.execute(outputs_schema)
+
```
dmalan: blank line

model.py
```diff
@@ -0,0 +1,42 @@
...
+    
+    outputs_schema = '''
+    CREATE TABLE IF NOT EXISTS outputs (
+        id BIGINT NOT NULL AUTO_INCREMENT,
+        input_id BIGINT,
+        output TEXT,
+        PRIMARY KEY (id),
+        FOREIGN KEY (input_id) REFERENCES inputs(id) ON DELETE CASCADE
+    ) ENGINE=InnoDB;
+    '''
```
dmalan: Is this the best approach or is there a de facto standard (or at least popular) "migration" library we should be using, in anticipation of making future table changes?

model.py
brianyu28: For some reason, it seems to be; even with `IF NOT EXISTS` present in the code, `MySQLdb` still prints out `table already exists` warnings every time `configure()` is executed unless the warnings are suppressed.

dmalan: k

dmalan:
I've created the server-side DB and credentials, fyi. But will wait before merging (re question about migration).

model.py
brianyu28: Though I haven't tried it before, looks like SQLAlchemy (http://www.sqlalchemy.org/) is a fairly popular library. It doesn't seem to support migrations natively, but there do appear to be extensions of it (https://pypi.python.org/pypi/alembic) which do.

dmalan: Hm, okay, want to poke around a bit more to see what most folks tend to use? Use case is... if we make changes to the DB tables down the road, ideally we just want the server to run a "migration," essentially a set of `ALTER` commands that updates the schema automatically.

model.py
```diff
@@ -0,0 +1,42 @@
...
+# disable 'table already exists' warnings
+filterwarnings('ignore', category=mdb.Warning)
+
+def configure():
+    inputs_schema = '''
+    CREATE TABLE IF NOT EXISTS inputs (
+        id BIGINT NOT NULL AUTO_INCREMENT,
+        cmd VARCHAR(1024) NULL,
+        script TEXT NOT NULL,
+        username VARCHAR(32) NULL,
```
dmalan: PS let's add an `INDEX` on `username` and on `created` so we can implement hearts more efficiently (by querying for users' past usage).

dmalan:
Just let me know when ready, holding off on merging for now.

brianyu28:
Closing because using a migration library instead in PR #14

PR #14, opened 2016-08-28
MySQL Configuration with Alembic

brianyu28:
Think this should work.

New setup for the database configuration works as follows:

- Interaction with the database is done through the `SQLAlchemy` library, which is similar to the Python MySQL library we were using previously, except it allows us to represent tables as Python classes (found in `manage.py`). `Alembic` is a library which adds migration capabilities to `SQLAlchemy`, which is connected to Flask via the `Flask-Migrate` library.

- Migrations are stored in the `migrations` directory. Any time a change is made to the database structure by editing the classes in `manage.py`, we can run `python manage.py db migrate` to generate new migration files. Each time the application is deployed, `flask_migrate.upgrade()` is called, and applies new migrations to the database.


Dockerfile
```diff
@@ -3,3 +3,8 @@ EXPOSE 8080
 
 # for development
 ENV PATH /srv/www/opt/cs50/help50/bin:"$PATH"
+
+# dependencies
+RUN apt-get update && apt-get install -y libmysqlclient-dev
+RUN pip3 install mysqlclient 
+RUN pip3 install mysqlclient Flask-Migrate Flask-SQLAlchemy
```
dmalan: Looks like `mysqlclient` is duplicated?

application.py
```diff
@@ -1,13 +1,26 @@
 from flask import abort, Flask, render_template, request
 import helpers
+import manage
 import re
+from flask_sqlalchemy import SQLAlchemy
+from flask_migrate import Migrate
+import flask_migrate
```
dmalan: Le's move these `from` lines up top alphabetically. Easier to skim for imports.

dmalan: And this `import`.

migrations/alembic.ini
```diff
@@ -0,0 +1,45 @@
...
+
+[handler_console]
+class = StreamHandler
+args = (sys.stderr,)
+level = NOTSET
+formatter = generic
+
+[formatter_generic]
+format = %(levelname)-5.5s [%(name)s] %(message)s
+datefmt = %H:%M:%S
```
dmalan: Is this file definitely necessary? Can these defaults not be assumed? Seems like a lot of detail to have to provide per app.

migrations/env.py
```diff
@@ -0,0 +1,87 @@
+from __future__ import with_statement
+from alembic import context
+from sqlalchemy import engine_from_config, pool
+from logging.config import fileConfig
+import logging
```
dmalan: Sorting here too.

migrations/env.py
```diff
@@ -0,0 +1,87 @@
...
+    try:
+        with context.begin_transaction():
+            context.run_migrations()
+    finally:
+        connection.close()
+
+if context.is_offline_mode():
+    run_migrations_offline()
+else:
+    run_migrations_online()
```
dmalan: Hm, definitely necessary? This is feeling complex just to do something cookie-cutter like migrations.

migrations/script.py.mako
```diff
@@ -0,0 +1,22 @@
...
+from alembic import op
+import sqlalchemy as sa
+${imports if imports else ""}
+
+def upgrade():
+    ${upgrades if upgrades else "pass"}
+
+
+def downgrade():
+    ${downgrades if downgrades else "pass"}
```
dmalan: This too?

manage.py
```diff
@@ -0,0 +1,34 @@
...
+app.config['SQLALCHEMY_DATABASE_URI'] = "mysql://" + os.environ["MYSQL_USERNAME"] + ":" + os.environ["MYSQL_PASSWORD"] + "@" + os.environ["MYSQL_HOST"] + "/" + os.environ["MYSQL_DATABASE"]
+app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = True
+
+db = SQLAlchemy(app)
+migrate = Migrate(app, db)
+
+manager = Manager(app)
+manager.add_command('db', MigrateCommand)
+
+class Input(db.Model):
```
dmalan: Should these be here, given that they're in `ad05a8ee42fc_.py` now?

brianyu28:
I should have clarified: all of the files in `migrations` are generated automatically. I didn't actually write `alembic.ini`, `env.py`, `script.py.mako`, etc., those were automatically generated by Alembic.

migrations/versions/ad05a8ee42fc_.py
```diff
@@ -0,0 +1,45 @@
+"""empty message
```
dmalan: Can/should this file be named something like `YYYY-MM-DD` so that they sort in a directory and it's easier to see which came first?

dmalan:
Re the files in `migrations`, are they all strictly necessary to commit? Surprised there's so much bulk is all. Migration libraries I've seen in node.js and PHP tend just to require the migration file itself, and you run some migration command that's simply a dependency.

manage.py
brianyu28: The classes in `manage.py` need to exist to facilitate future migrations. If we need to make changes to the database structure, we can edit the classes in `manage.py`, then use `python manage.py db migrate`, which will generate a new file in `migrations/versions/` containing the instructions for the `ALTER TABLE` commands that will modify the database structure appropriately. `ad05a8ee42fc_.py` is just the file for the first migration, but as the database structure changes, it will eventually become outdated.

migrations/versions/ad05a8ee42fc_.py
brianyu28: Alembic allows naming of migrations, but seems to want to keep the identifier of the migration at the start of the filename (when I tried it, it generated a file called `ae1a7f8e342f_2016_08_29.py` instead). I'll experiment with this and see if I can eliminate the need for the identifier to be in the filename.

brianyu28:
I tried removing the files in `migrations`, and it seems that they're all necessary in order to successfully perform and apply migrations.

According to Alembic's documentation (http://alembic.zzzcomputing.com/en/latest/tutorial.html#creating-an-environment), the files form the "Migration Environment" which is used to maintain migrations. The scripts in `migrations` are invoked by the migration command, and I would guess that the scripts are generated in the project filesystem itself rather than in a dependency to make it easier to edit or customize.

dmalan:
Hey, just catching up... Looks like this is indeed ready for merge, @brianyu28?

brianyu28:
Yes, this is ready now.

PR #17, opened 2016-08-29
changed help50 to use bash -l so as to source aliases

dmalan:


brianyu28:
Hm, I tried replacing `/opt/cs50/bin/help50` in my IDE with this version, and then re-running `help50 mv foo bar` (after opening a new terminal window), and it still doesn't appear to be aliasing.

dmalan:
Sorry! Should've tested.

dmalan:
Try https://github.com/cs50/help50/pull/18!

PR #18, opened 2016-08-29
changed invocation of bash to interactive shell

dmalan:
Fixes bug whereby aliased commands like `rm -i` and executed by `help50` as just `rm`.


brianyu28:
Aside from the issue pointed out in #19, this works for me.

PR #22, opened 2016-09-05
update db with output_id and reviewed

brianyu28:
I've deleted the old initial migration and replaced it with this one, which has the following changes:

- `reviewed` field has been added to `inputs`, with default value `False`.
- `input_id` has been removed from `outputs`, replaced by `output_id` in `inputs`, which is nullable.

Per your Asana comment, probably best to get rid of the current server-side tables before merging, such that the new initial migration can replace them.

I believe @annieechen is working on logging the inputs/outputs.


manage.py
```diff
@@ -16,19 +16,20 @@
...
     script = db.Column(db.Text, nullable=False)
     username = db.Column(db.String(length=32), nullable=True, index=True)
     created = db.Column(db.DateTime, default=func.now(), index=True)
-
-class Output(db.Model):
-    __tablename__ = "outputs"
-    id = db.Column(db.BigInteger, primary_key=True, nullable=False, autoincrement=True)
-    input_id = db.Column(db.BigInteger, db.ForeignKey("inputs.id"))
-    output = db.Column(db.Text)
+    output_id = db.Column(db.BigInteger, db.ForeignKey("outputs.id"), nullable=True)
```
dmalan: Does declaring this a foreign key automatically create its index?

manage.py
```diff
@@ -16,19 +16,20 @@
...
+    reviewed = db.Column(db.Boolean, default=False)
```
dmalan: This should be indexed too so we can search on it more efficiently.

migrations/versions/3b25e13c81bf_.py
```diff
@@ -16,30 +16,31 @@
...
-    sa.Column('input_id', sa.BigInteger(), nullable=True),
-    sa.Column('output', sa.Text(), nullable=True),
-    sa.ForeignKeyConstraint(['input_id'], ['inputs.id'], ),
-    sa.PrimaryKeyConstraint('id')
-    )
     ### end Alembic commands ###
 
 
 def downgrade():
     ### commands auto generated by Alembic - please adjust! ###
```
dmalan: Comment outdated.

migrations/versions/3b25e13c81bf_.py
```diff
@@ -16,30 +16,31 @@
 
 def upgrade():
     ### commands auto generated by Alembic - please adjust! ###
```
dmalan: Comment outdated.

manage.py
brianyu28: Yeah, InnoDB with MySQL requires that all foreign keys be indexed.

brianyu28:
Made adjustments in 03079c8

PR #25, opened 2016-09-05
log inputs/outputs and add /review

brianyu28:


application.py
```diff
@@ -50,15 +51,40 @@ def index():
...
+        if ("password" in request.form):
+            if (request.form.get("password") == os.environ["HELP50_PASSWORD"]):
+                return render_template("review.html", inputs=model.unreviewed_matchless())
+            else:
+                return render_template("review_auth.html", invalid=True)
+            
+        # user submitted form on review page
+        else:
+            for input_id in request.form:
+                model.mark_reviewed(input_id)
```
dmalan: nbd, but is there a way to do bulk updates, rather than one SQL query per row?

model.py
```diff
@@ -0,0 +1,38 @@
...
+def log(cmd, username, script, output):
+    output_id = None
+    # create output
+    if (output):
+        help_out = Output(output=output)
+        session.add(help_out)
+        session.commit()
+        output_id = help_out.id
+    
+    help_in = Input(cmd=cmd, username=username, script=script, output_id=output_id, created=datetime.now(timezone('America/New_York')))
```
dmalan: Generally best to store times in UTC.

templates/review.html
```diff
@@ -0,0 +1,31 @@
+{% extends "layout.html" %}
+{% block body %}
+help50 Review
+<br/>
+<form action="/review/" method="post">
```
dmalan: Just `/review` suffices since not a directory?

templates/review.html
```diff
@@ -0,0 +1,31 @@
...
+            <th>timestamp</th>
+            <th>script</th>
+        </tr>
+        {% for input in inputs %}
+        <tr>
+            <td><input type="checkbox" name={{ input.id }} value=true></td>
+            <td>{{ input.id }}</td>
+            <td>{{ input.cmd }}</td>
+            <td>{{ input.created }}</td>
+            <td>{{ input.script }}</td>
```
dmalan: How will this render if multiple lines? Should be monospaced too? Perhaps collapsed by default so that only one line displays until clicked?

templates/review_auth.html
```diff
@@ -0,0 +1,14 @@
+{% extends "layout.html" %}
+{% block body %}
+<form action="/review/" method="post">
```
dmalan: `/review`

application.py
brianyu28: Wasn't able to find a great way to do this after a first pass through the documentation, but I'll look into this more for as a possible improvement.

dmalan:
Just let me know when you think ready to merge, and I'll watch the DB as we test!

templates/review.html
brianyu28: Ends up looking like this:

screen shot 2016-09-05 at 7 22 04 pm

brianyu28:
I'd say this is ready— I can work on improving the UI of the review page (like adding collapsibility) next, but this should be sufficient to start logging any "Sorry"s so that I can track them and work on adding more matchers.

templates/review.html
dmalan: Nice.

dmalan:
Sounds great.

PR #29, opened 2016-09-06
successful compile message for make

brianyu28:
Cases like input `10` and `15` on `/review`, seem to be the result of a situation where someone ran `help50` on a program that compiles successfully. In that case, right now `help50` gives a `Sorry!` message. This update, which checks to see if the `script` is just a one-line `clang` message, handles that case.


brianyu28:
I'll mark `10` and `15` as reviewed on `/review`.

brianyu28:
Nevermind, looks like you're one step ahead of me.

Issue #23, opened 2016-09-05
expected parameter declarator


dmalan:
**input:**

```
#include <stdio.h>

int main(void);

printf("hello, world\n");
```

**output:**

```
$ clang bug1.c
bug1.c:5:8: error: expected parameter declarator
printf("hello, world\n");
       ^
bug1.c:5:8: error: expected ')'
bug1.c:5:7: note: to match this '('
printf("hello, world\n");
      ^
bug1.c:5:1: warning: type specifier missing, defaults to 'int' [-Wimplicit-int]
printf("hello, world\n");
^~~~~~
bug1.c:5:1: error: conflicting types for 'printf'
/usr/include/stdio.h:362:12: note: previous declaration is here
extern int printf (const char *__restrict __format, ...);
           ^
1 warning and 3 errors generated.
```

**current help but not right:**

```
Make sure that all opening parentheses `(` are matched with a closing parenthesis `)` in bug1.c.
In particular, check to see if you are missing a closing parenthesis on line 5 of bug1.c.
```

PR #30, opened 2016-09-06
match 'expected parameter declarator'

brianyu28:
There are two main ways I've seen to generate the `expected parameter declarator` error. Either the user has code like the example in Issue #23 with `main(void);` followed by code not embraced in curly braces, or the user has a function or prototype with a value in the parentheses. For instance, a prototype like:

```
int square(28);
```

generates the same error. There's no good way for us to tell which case caused it, so this output just provides explanations for both scenarios.

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -92,8 +92,8 @@ def help(lines):
     matches = re.search(r"^([^:]+):(\d+):\d+: error: expected parameter declarator", lines[0])
     if matches:
         after = [
-            "If you're trying to call a function on line {} of `{}`, be sure that it's being called inside of curly braces within a function. Also check that the function's header (the line introducing the function's name) doesn't end in a semicolon.".format(matches.group(2), matches.group(1)),
-            "Alternatively, if you're trying to define a function or prototype on line {} of `{}`, be sure each argument to the function is formatted as a data type followed by a variable name.".format(matches.group(2), matches.group(1))
+            "If you're trying to call a function on line {} of `{}`, be sure that you're calling it inside of curly braces within a function. Also check that the function's header (the line introducing the function's name) doesn't end in a semicolon.".format(matches.group(2), matches.group(1)),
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -83,7 +83,20 @@ def help(lines):
...
+    # $ clang foo.c
+    # /tmp/foo-1ce1b9.o: In function `main':
+    # foo.c:3:12: error: expected parameter declarator
+    # int square(28);
+    #            ^
+    matches = re.search(r"^([^:]+):(\d+):\d+: error: expected parameter declarator", lines[0])
+    if matches:
+        after = [
+            "If you're trying to call a function on line {} of `{}`, be sure that it's being called inside of curly braces within a function. Also check that the function's header (the line introducing the function's name) doesn't end in a semicolon.".format(matches.group(2), matches.group(1)),
+            "Alternatively, if you're trying to define a function or prototype on line {} of `{}`, be sure each argument to the function is formatted as a data type followed by a variable name.".format(matches.group(2), matches.group(1))
```
dmalan: s/define/declare/

To "define" means to implement.

helpers/clang.py
```diff
@@ -83,7 +83,20 @@ def help(lines):
...
+
+    # $ clang foo.c
+    # /tmp/foo-1ce1b9.o: In function `main':
+    # foo.c:3:12: error: expected parameter declarator
+    # int square(28);
+    #            ^
+    matches = re.search(r"^([^:]+):(\d+):\d+: error: expected parameter declarator", lines[0])
+    if matches:
+        after = [
+            "If you're trying to call a function on line {} of `{}`, be sure that it's being called inside of curly braces within a function. Also check that the function's header (the line introducing the function's name) doesn't end in a semicolon.".format(matches.group(2), matches.group(1)),
```
dmalan: "be sure that you're calling it inside..."

brianyu28:
Made the appropriate fixes.

dmalan:
Mind pulling and addressing merge conflicts?

brianyu28:
Addressed and should be good now.

PR #41, opened 2016-09-07
detect CS50 Library functions

brianyu28:


dmalan:
Mind adding `Get*` versions for now too?

brianyu28:
Added.

PR #44, opened 2016-09-07
preventing user from running help50 sans pipe and sans argv

dmalan:
@glennholloway, any concerns with this change? Realized that usage is confusing if student accidentally runs just `help50` sans arguments (and sans pipe). So this effectively disables that interactive feature (which seems unnecessary for this type of program).

CC @brianyu28

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 MAINTAINER = "CS50 <sysadmins@cs50.harvard.edu>"
 NAME = help50
-VERSION = 1.0.3
+VERSION = 1.1.0
 
 .PHONY: bash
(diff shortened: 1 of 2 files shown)


glennholloway:
@dmalan @brianyu28 Can't think of anything important this restriction obstructs. Without it the user could start `help50` and then copy and paste a transcript taken from somewhere else into the terminal (followed by ^D). But that's not an important use case, or one that students are likely to think of.

dmalan:
Indeed, that's what I originally envisioned, but feels like the downside of getting "stuck" in that mode outweighs the upside of the feature!

Issue #36, opened 2016-09-06
change all "error:" to "(?:warning|error):"?


dmalan:
Perhaps in case students omit `-Werror` from future `Makefile`s of their own?

PR #46, opened 2016-09-08
change all "error:" to "(?:warning|error):"

brianyu28:
Addressing #36

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -3,7 +3,7 @@ def help(lines):
 
     # $ clang foo.c
     # foo.c:13:25: error: adding 'int' to a string does not append to the string [-Werror,-Wstring-plus-int]
-    matches = re.search(r"^([^:]+):(\d+):\d+: error: adding '(.+)' to a string does not append to the string", lines[0])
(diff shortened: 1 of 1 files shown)


dmalan:
Can all of these definitely be either warnings or errors? Or are some only errors, irrespective of `-Werror`?

brianyu28:
The ones that can be warnings or errors should already be marked as `(?:warning|error):`, even prior to this PR. I'm fairly certain that all of these changes included here are ones that are only `error`.

When you posted the issue, I thought you meant that for consistency, we should turn everything into `(?:warning|error):` so we wouldn't have to worry now or in the future about checking to see whether a given message could be either one. If I misunderstood, then we can probably just close this and continue with the existing workflow.

dmalan:
Oh I just meant we should handle both _if_ we might see the message as either a warning or an error, depending on flags. Feels bad to add "warning|" to the regexes if we'll never encounter them as warnings.

brianyu28:
Makes sense, I agree. Yeah, the ones that can be warnings are already marked as `(?:warning|error):`. Sorry for the misinterpretation, closing this now.

PR #47, opened 2016-09-08
add matcher for semicolon after do/while

brianyu28:
The `expected ';' after do/while statement` has come up 11 times in `/review` (from a common error in `mario.c` and `water.c`), so I updated the existing missing semicolon matcher to handle this case.

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -158,9 +158,9 @@ def help(lines):
     #    printf("hello, world!")
     #                           ^
     #                           ;
-    matches = re.search(r"^[^:]+:(\d+):\d+: error: expected ';' (?:after\sexpression|at\send\sof\sdeclaration)", lines[0])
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -158,9 +158,9 @@ def help(lines):
     #    printf("hello, world!")
     #                           ^
     #                           ;
-    matches = re.search(r"^[^:]+:(\d+):\d+: error: expected ';' (?:after\sexpression|at\send\sof\sdeclaration)", lines[0])
+    matches = re.search(r"^([^:]+):(\d+):\d+: error: expected ';' (?:after\sexpression|at\send\sof\sdeclaration|after\sdo\/while\sstatement)", lines[0])
```
dmalan: Why the `\s` instead of ` `?

brianyu28: For some reason, I thought I recalled there being a problem with spaces separating words regarding how the `|` operator interpreted them, but I just tested with spaces and that works too. Changed to spaces.

PR #48, opened 2016-09-08
add match for empty if statement

annieechen:
for 311, 663, 668, 669

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -400,3 +400,14 @@ def help(lines):
             "Be sure to assign a value to `{}` before trying to access its value.".format(matches.group(3))
         ]
         return (lines[0:1], after)
+    
+    # $ clang foo.c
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -400,3 +400,14 @@ def help(lines):
             "Be sure to assign a value to `{}` before trying to access its value.".format(matches.group(3))
         ]
         return (lines[0:1], after)
+    
+    # $ clang foo.c
+    # /tmp/foo-1ce1b9.o: In function `main':
+    # foo.c:12:15: error: if statement has empty body [-Werror,-Wempty-body]
+    #   if (n > 0);
+    #             ^
+    matches = re.search(r"if statement has empty body", lines[0])
```
dmalan: Actually, do add detection of filename, line number, and (warning|error) here. Mind re-PRing?

PR #49, opened 2016-09-08
clarified "no rule to make target" helper

brianyu28:
Addressing @kzidane's comment here:

https://github.com/cs50/help50/commit/ce93e59f6f410e8ba6bafb42c77828ad6192ede7#commitcomment-18950855


dmalan:
:shipit:

PR #50, opened 2016-09-08
refine empty body matcher

annieechen:
also fixed a typo on the matcher above

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -395,7 +395,7 @@ def help(lines):
     matches = re.search(r"^([^:]+):(\d+):\d+: (?:warning|error): variable '(.*)' is uninitialized when used here", lines[0])
     if matches:
         after = [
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -406,8 +406,9 @@ def help(lines):
     # foo.c:12:15: error: if statement has empty body [-Werror,-Wempty-body]
     #   if (n > 0);
     #             ^
-    matches = re.search(r"if statement has empty body", lines[0])
+    matches = re.search(r"^([^:]+):(\d+):\d+: error: (if statement|while loop|for loop) has empty body", lines[0])
```
dmalan: Can this sometimes be a `warning` (without `-Werror` on)?

annieechen: Yep, addressed in commit 538e699

dmalan: ah ty

Issue #54, opened 2016-09-09
HTTP 500 occasionally


dmalan:
This error starts to recur:

```
App 116 stderr: ERROR [application] Exception on / [POST]
App 116 stderr: Traceback (most recent call last):
App 116 stderr:   File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1988, in wsgi_app
App 116 stderr:     response = self.full_dispatch_request()
App 116 stderr:   File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1641, in full_dispatch_request
App 116 stderr:     rv = self.handle_user_exception(e)
App 116 stderr:   File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1544, in handle_user_exception
App 116 stderr:     reraise(exc_type, exc_value, tb)
App 116 stderr:   File "/usr/local/lib/python3.4/dist-packages/flask/_compat.py", line 33, in reraise
App 116 stderr:     raise value
App 116 stderr:   File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1639, in full_dispatch_request
App 116 stderr:     rv = self.dispatch_request()
App 116 stderr:   File "/usr/local/lib/python3.4/dist-packages/flask/app.py", line 1625, in dispatch_request
App 116 stderr:     return self.view_functions[rule.endpoint](**req.view_args)
App 116 stderr:   File "/srv/www/application.py", line 56, in index
App 116 stderr:     model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), after)
App 116 stderr:   File "/srv/www/model.py", line 20, in log
App 116 stderr:     session.commit()
App 116 stderr:   File "/usr/local/lib/python3.4/dist-packages/sqlalchemy/orm/session.py", line 801, in commit
App 116 stderr:     self.transaction.commit()
App 116 stderr:   File "/usr/local/lib/python3.4/dist-packages/sqlalchemy/orm/session.py", line 390, in commit
App 116 stderr:     self._assert_active(prepared_ok=True)
App 116 stderr:   File "/usr/local/lib/python3.4/dist-packages/sqlalchemy/orm/session.py", line 214, in _assert_active
App 116 stderr:     % self._rollback_exception
App 116 stderr: sqlalchemy.exc.InvalidRequestError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction
[...]

PR #55, opened 2016-09-09
address illegal characters before input into database

brianyu28:
Is there a more elegant way to do this? This strips out any illegal characters before database input, although will take curly quotes (single and double) and convert to straight quotation marks.

Addressing #54

### DIFF
--- a/model.py
+++ b/model.py
@@ -11,15 +11,19 @@ engine = create_engine(uri)
 Session = sessionmaker(bind=engine)
 session = Session()
 
+# cleans text for db input
+def clean(text):
+    return text.replace(u"\u2018", "'").replace(u"\u2019", "'").replace(u"\u201C", "\"").replace(u"\u201D", "\"").encode("latin-1", "ignore")
+
 # logs input (and any helpful output)
 def log(cmd, username, script, output):
     output_id = None
     if (output):
-        help_out = Output(output=output)
+        help_out = Output(output=clean(output))
         session.add(help_out)
         session.commit()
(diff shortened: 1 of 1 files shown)


dmalan:
For something like this, it's most certainly a common issue, since arbitrary database inputs shouldn't cause such fatal failures. And definitely resist the temptation to reinvent a wheel that must exist, a la https://github.com/cs50/help50/pull/55/commits/09274d755d50ba6198cb252f155e15ef36a2dfff#diff-93ff136fae812392eb0f68d1ce89b7feR15.

I Googled `UnicodeEncodeError: 'latin-1' codec can't encode character`, which led me to http://stackoverflow.com/a/3943139, which does seem to explain. So I went ahead and converted our tables to `utf8`, per the below. Sounds like this _should_ fix!

```sql
ALTER TABLE alembic_version  CONVERT TO CHARACTER SET utf8 COLLATE utf8_unicode_ci;
ALTER TABLE inputs CONVERT TO CHARACTER SET utf8 COLLATE utf8_unicode_ci;
ALTER TABLE outputs CONVERT TO CHARACTER SET utf8 COLLATE utf8_unicode_ci;
```

dmalan:
Just to close the loop, seems this may have fixed earlier INSERTs, but for new INSERTs, these changes weren't sufficient (cf. http://stackoverflow.com/a/9157305). But https://github.com/cs50/help50/commit/3abe2e63b3d267e8a94581173fbfec62b51f966b should now address.

Issue #51, opened 2016-09-09
show only a subset of input along with output?


dmalan:
So I've been thinking about whether there's a better way to display `help50`'s output. At the moment, we display the entirety of its input, below which, in yellow, is its output. 

But I wonder if we should start showing only the first _n_ lines of input, where _n_ - 1 is the (last) line we matched. That'd shorten total output in many cases. And we could also start to highlight substrings in the input (e.g., filenames, line numbers, portions of error messages). 

Separately, we currently output lines of output, adding `\n`s irrespective of screen width. Am thinking we should start outputting "paragraphs" instead and let them wrap naturally.

Thoughts? Can we produce better output?

brianyu28:
Alternatively, we could adopt the web interface's model, which only displays the lines of code that we output in the `before` part of the `(before, after)` return value of any `help` function?

When I input:

```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    test.c  -lcs50 -lm -o test
test.c:6:8: error: expected '(' after 'if'
    if x == 28
       ^
1 error generated.
make: *** [test] Error 1
```

into the web interface, the resulting page I get just has the single line:

```
test.c:6:8: error: expected '(' after 'if'
```

followed by the `after` advice. I think there's definitely pedagogical value in showing less input, so that students can better link the semantics of the error message itself with `help50`'s interpretation and rely on it less in the future.

However, right now, if compiling a program results in three errors, and `help50` only catches the third one, then the command line interface would show all three errors, while the web interface would only show the third one. Is there a preferable option here?

dmalan:
Hm, something tells me we should show lines 0 through _n_ - 1, lest we hide the actual source of a problem (that we just haven't happened to match on yet)? 

Rather than have a matcher return `(before, after)`, should we instead have it return `(n, help)`, where `n` is the number of lines matched and `help` is just a renaming of `after`? 

I was hoping there might be a cleaner way, whereby we could just pass each helper the whole array but rely on an internal cursor, but seems https://docs.python.org/2/library/stdtypes.html#iterator-types, at least, doesn't offer much more than `next()`.

brianyu28:
Okay, I agree, that makes sense. `(n, help)` seems like a good way to deal with this. For cases like:

```
foo.c:5:15: error: expected ';' at end of declaration
    int x = 28
              ^
              ;
1 error generated.
make: *** [foo] Error 1
```

[...]

PR #57, opened 2016-09-10
Show a subset of input with output

brianyu28:
Addressing #51.

Each helper function now returns `(before, after)`, where `before` is the number of lines of the input (starting at `lines[0]`) that should be displayed. Any lines before `lines[0]` that weren't matched will also be displayed in the template.

If merged, this will (as of now) only change the web interface experience. To change the command-line experience as well, we'd then also have to change the `helpful.txt` template to display both `before` and `after` (with `after` in a different color, ideally), and then modify the `help50` script to not automatically display the output of `clang`, `make`, etc. and to not color all of the response from `help.cs50.net` (since `before` shouldn't be colored the same).

I didn't want to modify `helpful.txt` just yet, since otherwise students working right now would end up seeing two sets of error messages each time they run `help50`.

### DIFF
--- a/application.py
+++ b/application.py
@@ -54,7 +54,7 @@ def index():
                 if help:
                     after = "\n".join(help[1])
                     model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), after)
-                    return render_template("helpful." + format, before="\n".join(help[0]), after=after)
+                    return render_template("helpful." + format, before="\n".join(lines[:i+help[0]]), after=after)
 
         # unhelpful response
         model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), None)
--- a/application.py
+++ b/application.py
@@ -53,7 +53,7 @@ def index():
                 # helpful response
                 if help:
                     n, response = help
-                    response = "\n".join(response)
+                    response = " ".join(response)
                     model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), response)
                     return render_template("helpful." + format, before="\n".join(lines[:i+n]), after=response)
 
--- a/application.py
+++ b/application.py
@@ -52,9 +52,10 @@ def index():
 
                 # helpful response
                 if help:
-                    after = "\n".join(help[1])
-                    model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), after)
-                    return render_template("helpful." + format, before="\n".join(lines[:i+help[0]]), after=after)
+                    n, response = help
+                    response = "\n".join(response)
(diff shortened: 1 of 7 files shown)


dmalan:
Given that `before` and `after` are rather apple and orange now, with the former an `int` and the latter a `str`, should we perhaps change 

https://github.com/cs50/help50/pull/57/commits/924fb50d418ef896c3dc4e808282a13bcb7a6a23#diff-925fb778c109964ac347b28b014882cbR51 

to something like

```py
n, help = helpers.__dict__.get(helper).help(lines[i:])
```

and drop the `before` and `after` notion semantically?

brianyu28:
Makes sense to me, just committed some changes.

I kept the `before` and `after` terminology just for the template files, since in that case they're both lines of text (unless you have a better name in mind to change `before` to).

Also, storing the result of `help` into `n, help` runs into problems in the event that `help` returns `None`, so that separation happens a bit later. In the helper files, I replaced `after` with `response` (was thinking about using `help`, but didn't want to confuse it with the name of the function itself, which is also `help`), though let me know if you have a preferred name for that.

helpers/bash.py
```diff
@@ -5,38 +5,38 @@ def help(lines):
...
-        return (1, after)
+        return (1, response)
 
     # $ cd foo
     # bash: cd: foo: No such file or directory
     matches = re.search(r"^bash: (.+): No such file or directory", lines[0])
     if matches:
-        after = [
+        response = [
             "Are you sure `{}` exists".format(matches.group(1)),
```
dmalan: Missing question mark (just noticed).

application.py
```diff
@@ -52,9 +52,10 @@ def index():
 
                 # helpful response
                 if help:
-                    after = "\n".join(help[1])
-                    model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), after)
-                    return render_template("helpful." + format, before="\n".join(lines[:i+help[0]]), after=after)
+                    n, response = help
+                    response = "\n".join(response)
```
dmalan: Shall we join with ` ` now instead, to avoid some lines wrapping and some breaking, depending on terminal size?

helpers/bash.py
brianyu28: Fixed.

application.py
brianyu28: Sure, that makes sense.

dmalan:
@brianyu28 in anticipation of `join`ing on ` ` instead of `\n`, https://github.com/cs50/help50/pull/61.

dmalan:
:shipit:

Issue #56, opened 2016-09-10
change migration to create tables with UTF-8 charset and collation


dmalan:
A la https://github.com/cs50/help50/pull/55#issuecomment-246085722.

PR #59, opened 2016-09-10
utf8 migration

brianyu28:
Addressing #56, changes the default charset of the database to `utf8`. Alembic seems to have a known issue where certain changes [aren't detected](https://bitbucket.org/zzzeek/alembic/issues/180/detect-changes-in-type-parameters), so I had to manually add some lines to the migration file (source in comments).

I also found a recommendation in the SQLAlchemy [documentation](http://docs.sqlalchemy.org/en/latest/faq/sessions.html#this-session-s-transaction-has-been-rolled-back-due-to-a-previous-exception-during-flush-or-similar) to call `session.rollback()` if a database operation fails, so that the application doesn't continue trying to execute the failing operation in future operations. This should make it so that, if for some reason a user runs into a failed input, subsequent users aren't affected.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -18,11 +18,15 @@ manager.add_command('db', MigrateCommand)
 
 class Output(db.Model):
     __tablename__ = "outputs"
+    mysql_default_charset = "utf8",
+    mysql_collate = "utf8_general_ci"
     id = db.Column(db.BigInteger, primary_key=True, nullable=False, autoincrement=True)
     output = db.Column(db.Text)
 
 class Input(db.Model):
     __tablename__ = "inputs"
+    mysql_default_charset = "utf8",
+    mysql_collate =  "utf8_general_ci"
     id = db.Column(db.BigInteger, primary_key=True, nullable=False, autoincrement=True)
     cmd = db.Column(db.String(length=1024), nullable=True)
     script = db.Column(db.Text, nullable=False)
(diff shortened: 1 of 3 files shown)


manage.py
```diff
@@ -18,11 +18,15 @@
...
     __tablename__ = "outputs"
+    mysql_default_charset = "utf8",
+    mysql_collate = "utf8_general_ci"
     id = db.Column(db.BigInteger, primary_key=True, nullable=False, autoincrement=True)
     output = db.Column(db.Text)
 
 class Input(db.Model):
     __tablename__ = "inputs"
+    mysql_default_charset = "utf8",
+    mysql_collate =  "utf8_general_ci"
```
dmalan: These are instance variables defined in `db.Model` I assume? All we need do is set them?

model.py
```diff
@@ -32,4 +36,8 @@ def mark_reviewed(input_id):
     row = session.query(Input).filter(Input.id==input_id).first()
     if (row != None):
         row.reviewed = True
-    session.commit()
+    try:
+        session.commit()
+    except:
+        session.rollback()
+        raise
```
dmalan: What's going to happen if this `raise` happens? Will the user see the error?

manage.py
brianyu28: Yes, per http://docs.sqlalchemy.org/en/latest/dialects/mysql.html#create-table-arguments-including-storage-engines. The only problem is the Alembic's migration system didn't actually detect that I had made those changes to `manage.py`, which is why I had to manually add the commands to the migration file.

model.py
brianyu28: The user gets the same `500` page they would have seen otherwise. Subsequent users are the ones affected by this change: while previously they would have also gotten `500` errors (even for good requests), now they shouldn't.

dmalan:
:shipit:

PR #61, opened 2016-09-10
folding long lines, breaking on words; adding 'Asking for help...'

dmalan:
Uses `fold` to ensure that long messages from server wrap nicely on terminal window, given its current width. Adds a "Asking for help..." progress message in case server is slow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,6 @@ deb:
 	--deb-no-default-config-files \
 	--depends coreutils \
 	--depends curl \
-	--depends ncurses-bin \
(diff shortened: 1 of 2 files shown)


dmalan:
@glennholloway any concerns with https://github.com/cs50/help50/pull/61/files#diff-5306b6a56f5cfe829bb55cd7d8c6eb77, particularly `fold` and `$COLUMNS`?

dmalan:
@glennholloway just wanted to check in before we merge!

Issue #58, opened 2016-09-10
No help50 message for runtime "Floating point exception"


adamzu:
Could we add a useful message for such an error? Maybe something like: "Are you performing any possibly illegal mathematical operations (i.e. dividing by zero)?"

dmalan:
@adamzu can you add sample code that triggers that error? And on what platform did you execute it? I get the below on Ubuntu 14.04, for instance:

```c
#include <stdio.h>

int main(void)
{
    printf("%f\n", 1.0 / 0);
}
```

```
$ clang foo.c
$ ./a.out
inf
```

brianyu28:
This should generate it:

```
#include <stdio.h>

int main(void)
{
    int x = 0;
    printf("%d\n", 28 % x);
}
```

We don't currently have any matchers for runtime errors. @dmalan, will `help50` be able to process runtime error output even if the program has previously prompted the user for input? Also, thoughts on what file to put these in? Our convention of `foo.py` where `foo` is the name of the program generating the error likely won't work here.

adamzu:
yeah I did something like this:
`int main(void)
{
    int x = 24;
    for(int i = x; i >= 0; i--){
        int quotient = x / i;
        printf("%d/n", quotient);
    }
}`

dmalan:
Thanks. Yup, `help50` supports interactive use. E.g.:

```c
#include <cs50.h>
#include <stdio.h>

int main(void)
{
    printf("int: ");
    int x = get_int();
    printf("%d\n", 28 % x);
}
```

```
$ make foo
$ help50 ./foo
int: 0
Sorry, not sure how to help!
```

@brianyu28 maybe just call this matcher `a.out.py` or such, so we remember it's generic?

adamzu:
yeah the compiler catches explicit divides by zero (i.e. int x = 5 / 0; – help50 doesn't have an interpretation of those either but the compiler error is pretty self-explanatory), but not the implicit ones.

```
debug.c:16:11: error: division by zero is undefined [-Werror,-Wdivision-by-zero]
    x = 5 / 0;
          ^ ~
1 error generated.
```

PR #76, opened 2016-09-11
a.out helpers

brianyu28:
Addressing #58 and #60.

The name `a.out.py` ran into trouble when trying to import it into `application.py` because Python uses the `.` as a way to separate modules.

Tentatively renamed to `aout.py`, which I'm not convinced is the most elegant name for the file, thoughts on anything better?

### DIFF
--- a/helpers/aout.py
+++ b/helpers/aout.py
@@ -1,0 +1,27 @@
+import re
+def help(lines):
+    
+    # $ ./a.out
+    # Floating point exception
+    matches = re.match("Floating point exception", lines[0])
+    if matches:
+        response = [
+            "Looks like somewhere in your program, you're trying to divide a number by 0.",
+            "Check to see where in your program you're using the `/` or `%` operators, and be sure you never divide by 0 or calculate a number modulo 0.",
+            "If still unsure of where the problem is, stepping through your code with `debug50` may be helpful!"
+        ]
+        return (1, response)
+    
+    # $ ./a.out
+    # Segmentation fault
+    matches = re.match("Segmentation fault", lines[0])
+    if matches:
+        response = [
+            "Looks like your program is trying to access areas of memory that it isn't supposed to access.",
+            "There are many causes for segmentation faults, but you may want to consider:",
+            "Are you trying to access an element of an array beyond the size of the array?",
+            "Are you taking care to only dereference a pointer after you've initialized it (perhaps with `malloc`)?",
+            "Have you checked that any pointers you're dereferencing aren't NULL?",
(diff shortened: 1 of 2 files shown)


helpers/aout.py
```diff
@@ -0,0 +1,27 @@
...
+        ]
+        return (1, response)
+    
+    # $ ./a.out
+    # Segmentation fault
+    matches = re.match("Segmentation fault", lines[0])
+    if matches:
+        response = [
+            "Looks like your program is trying to access areas of memory that it isn't supposed to access.",
+            "There are many causes for segmentation faults, but you may want to consider:",
```
dmalan: Not sure we need this preamble, since many helpers already offer multiple suggestions?

helpers/aout.py
```diff
@@ -0,0 +1,27 @@
...
+            "Are you trying to access an element of an array beyond the size of the array?",
+            "Are you taking care to only dereference a pointer after you've initialized it (perhaps with `malloc`)?",
+            "Have you checked that any pointers you're dereferencing aren't NULL?",
```
dmalan: \`NULL\`

helpers/aout.py
```diff
@@ -0,0 +1,27 @@
...
+    
+    # $ ./a.out
+    # Segmentation fault
+    matches = re.match("Segmentation fault", lines[0])
+    if matches:
+        response = [
+            "Looks like your program is trying to access areas of memory that it isn't supposed to access.",
+            "There are many causes for segmentation faults, but you may want to consider:",
+            "Are you trying to access an element of an array beyond the size of the array?",
+            "Are you taking care to only dereference a pointer after you've initialized it (perhaps with `malloc`)?",
```
dmalan: This seems the opposite of what we want to say? "Are you dereferencing a pointer that you haven't initialized?"

helpers/aout.py
dmalan: I think we should focus on what the code is doing, not the student, so to speak. E.g., "Are you dereferencing a pointer whose value is \`NULL\`?"

helpers/aout.py
```diff
@@ -0,0 +1,27 @@
...
+            "Have you checked that any pointers you're dereferencing aren't NULL?",
+            "Are you trying to dereference a pointer after you've freed it?"
```
dmalan: s/trying to//

dmalan:
Not opposed to just `a.py`!

brianyu28:
Fixed.

dmalan:
:shipit:

PR #82, opened 2016-09-11
catch misspellings of "caesar"

brianyu28:
This catches when someone runs `make ceasar` and spells `caesar` wrong in the `make` command, or when someone runs `make caesar` but has spelled their `caesar.c` file wrong.

### DIFF
--- a/helpers/make.py
+++ b/helpers/make.py
@@ -15,6 +15,16 @@ def help(lines):
     # make: *** No rule to make target `foo'.  Stop.
     matches = re.search(r"^make: \*\*\* No rule to make target `(.+)'.  Stop.", lines[0])
     if matches:
+        if matches.group(1) in ["ceasar", "ceaser", "cesar", "caesaer"]:
+            response = [
+                "Did you mean to type `make caesar`?"
+            ]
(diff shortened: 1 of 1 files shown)


coltonoscopy:
This could be made very general-purpose using Python's popular fuzzywuzzy library for fuzzy string matching as follows:

1. install fuzzywuzzy and include it as part of help50 as a whole
2. compile a list of all the correctly spelled C programs used throughout the semester, caesar included
3. use fuzzywuzzy's "ratio" function, testing to see if the ratio between a user's spelling of their program and the correct spelling is > 70 or so, for each correctly spelled program
4. extract whichever is the highest rated above that threshold to be used in the help50 message

This would remove the need to hardcode potential misspellings and make it applicable to every C-based problem set!

Examples (including ratio): https://github.com/seatgeek/fuzzywuzzy

dmalan:
Not a bad principle in general, but tbh, `caesar` is a bit of an anomaly, given how frequently it's misspelled. `stdio.h` is perhaps another. @brianyu28 would this be helpful to swap in anywhere? Or wait until we see some other common misspellings?

PR #83, opened 2016-09-12
"array subscript is not an integer" matcher and extraction tools

brianyu28:
Addressing #63 and `/review` input `2146`.

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -25,6 +25,8 @@ def help(lines):
             response = [
                 "Looks like you're trying to access an element of the array `{}` on line {} of `{}`, but your index `{}` is not of type `int`.".format(array, matches.group(2), matches.group(1), index)    
             ]
+            if index.startswith("\"") and index.endswith("\""):
+                response.append("Right now, your index is of type `string` instead.")
         else:
             response = [
                 "Looks like you're trying to access an element of an array on line {} of `{}`, but your index is not of type `int`.".format(matches.group(2), matches.group(1))
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -1,4 +1,6 @@
 import re
+from tools import *
+
 def help(lines):
 
     # $ clang foo.c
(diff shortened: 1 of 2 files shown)


helpers/clang.py
```diff
@@ -10,6 +10,21 @@ def help(lines):
...
+    
+    # $ clang foo.c
+    # foo.c:13:25: error: adding 'int' to a string does not append to the string [-Werror,-Wstring-plus-int]
+    # foo.c:6:21: error: array subscript is not an integer
+    #     printf("%i\n", x["28"]);
+    #                     ^~~~~ 
+    matches = re.search(r"^([^:]+):(\d+):\d+: error: array subscript is not an integer", lines[0])
+    if matches:
+        response = [
+            "Looks like you're trying to access an element of an array on line {} of `{}`, but your index is not of type `int`.".format(matches.group(2), matches.group(1)),
```
dmalan: Can we potentially be more precise, extracting the array's name from `lines[2]` using the `^` on `lines[3]`, capturing any chars to the left of the `[` that are valid var names?

brianyu28: Sure. Code to do that feels pretty generalizable, I can envision other scenarios when we might also want to extract the name of an array based on the location of `^`. Think it'd be worth creating a file with some functions that take care of such extractions (extracting array names, as well as variable names, and possibly symbols), so that we can reuse that sort of functionality throughout `clang.py` where appropriate? (Looking ahead to unhandled error messages on `/reivew` like `variable length array declaration not allowed at file scope` and `array initializer must be an initializer list or string literal`).

dmalan: Sure, seems reasonable, if you think we're likely to use a few times. Though sometimes the `^` just points at syntax (e.g., a semicolon). Though maybe a (separate) function that just grabs everything over `~` could be handy too.

brianyu28:
Alright, added some extraction tools in a new file called `tools.py`.

`char_extract(lines)` will return the single character immediately above the `^`, so input like:

```
for (int i = 0, i < 28, i++)
                  ^
```

will return `<`.

`tilde_extract(lines)` returns everything above the `~` symbols, and `var_extract(lines)` returns the name of the variable pointed to by `^` (with an optional parameter for determining whether the `^` appears at the beginning of the variable name or at the end, as it appears for arrays).

So for this particular example, with an error output of:

```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    foo.c  -lcs50 -lm -o foo
foo.c:6:26: error: array subscript is not an integer
    printf("%i\n", height["28"]);
                         ^~~~~
1 error generated.
make: *** [foo] Error 1
```

we make use of `tilde_extract(lines)` and `var_extract(lines)` to generate the following output:

```
Looks like you're trying to access an element of the array `height` on line 6 of `foo.c`, but your index `"28"` is not of type `int`. Make sure your index (the value between square brackets) is an `int`.
```

helpers/clang.py
```diff
@@ -12,16 +14,22 @@ def help(lines):
...
+        index = tilde_extract(lines[1:3])
+        if array and index:
+            response = [
+                "Looks like you're trying to access an element of the array `{}` on line {} of `{}`, but your index `{}` is not of type `int`.".format(array, matches.group(2), matches.group(1), index)    
+            ]
+        else:
+            response = [
+                "Looks like you're trying to access an element of an array on line {} of `{}`, but your index is not of type `int`.".format(matches.group(2), matches.group(1))
+            ]
+        response.append("Make sure your index (the value between square brackets) is an `int`.")
```
dmalan: Nice. Can we go a step further here, and detect that the value is surrounded by double quotes so we can say "your index is of type `string` instead of `int`" or such?

dmalan:
:shipit:

helpers/clang.py
brianyu28: Just pushed a change to detect this: it'll work for a case when the user directly included a double-quoted string as an index. What seems more likely, though, (and which happened in the input this matcher is based on) is that a student would write `array[i]` where `i` is the name of some variable that is not an `int`. At that point, we don't really have a good way of determining the type of `i`.

dmalan: Ah, very true.

Issue #69, opened 2016-09-10
error: 'main' must return 'int'


dmalan:
```
hello.c:3:1: error: 'main' must return 'int'
void main()
^~~~
int
1 error generated.
```

thctamm:
https://github.com/cs50/help50/pull/86

PR #85, opened 2016-09-14
add "main must return int" matcher

brianyu28:
Addressing #69, and `/review` inputs 1430, 1612, and 2053.

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -380,7 +380,7 @@ def help(lines):
     # void main(void)
     # ^~~~
     # int
-    matches = re.search(r"^([^:]+):(\d+):\d+: error:", lines[0])
+    matches = re.search(r"^([^:]+):(\d+):\d+: error: 'main' must return 'int'", lines[0])
     if matches:
         response = [
             "Your `main` function (declared on line {} of `{}`) must have a return type `int`.".format(matches.group(2), matches.group(1))
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -23,7 +23,7 @@ def help(lines):
         index = tilde_extract(lines[1:3])
         if array and index:
             response = [
-                "Looks like you're trying to access an element of the array `{}` on line {} of `{}`, but your index `{}` is not of type `int`.".format(array, matches.group(2), matches.group(1), index)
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -375,6 +375,23 @@ def help(lines):
 
     # $ clang foo.c
     # /tmp/foo-1ce1b9.o: In function `main':
+    # foo.c:3:1: error: 'main' must return 'int'
+    # void main(void)
+    # ^~~~
+    # int
+    matches = re.search(r"^([^:]+):(\d+):\d+: error:", lines[0])
```
thctamm: The actual error message ('main' must return 'int') seems to be missing

brianyu28: Good catch, thanks.

helpers/clang.py
```diff
@@ -12,18 +12,18 @@ def help(lines):
...
-    #                     ^~~~~ 
+    #                     ^~~~~
     matches = re.search(r"^([^:]+):(\d+):\d+: error: array subscript is not an integer", lines[0])
     if matches:
         array = var_extract(lines[1:3], left_aligned=False)
         index = tilde_extract(lines[1:3])
         if array and index:
             response = [
-                "Looks like you're trying to access an element of the array `{}` on line {} of `{}`, but your index `{}` is not of type `int`.".format(array, matches.group(2), matches.group(1), index)    
+                "Looks like you're trying to access an element of the array `{}` on line {} of `{}`, but your index `{}` is not of type `int`.".format(array, matches.group(2), matches.group(1), index)
```
dmalan: Minor, but feels like index should be a parenthetical:

`but your index (\`{}\`) is`

helpers/clang.py
```diff
@@ -375,6 +375,23 @@ def help(lines):
...
+    # ^~~~
+    # int
+    matches = re.search(r"^([^:]+):(\d+):\d+: error:", lines[0])
+    if matches:
+        response = [
+            "Your `main` function (declared on line {} of `{}`) must have a return type `int`.".format(matches.group(2), matches.group(1))
+        ]
+        if len(lines) >= 3:
+            cur_type = var_extract(lines[1:3])
+            response.append("Right now, it has return type `{}`.".format(cur_type))
```
dmalan:
```
has a return type of `{}`
```

dmalan:
:shipit:

PR #87, opened 2016-09-14
add case for unused expression results

thctamm:
response for issue [68](https://github.com/cs50/help50/issues/68)

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -280,6 +280,18 @@ def help(lines):
             return (2, response)
 
     # $ clang foo.c
+    # foo.c:6:16: error: expression result unused [-Werror,-Wunused-value]
+    # n*12;
+    #  ^ 1 error generated.
+    matches = re.search(r"^([^:]+):(\d+):\d+: error: expression result unused", lines[0])
+    if matches:
+        response = [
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -278,7 +278,19 @@ def help(lines):
         if matches:
             response = ["The first argument to `{}` on line {} of `{}` should be a double-quoted string.".format(matches.group(1), line, file)]
             return (2, response)
-        
+
+    # $ clang foo.c
+    # foo.c:6:16: error: expression result unused [-Werror,-Wunused-value]
+    # n*12;
+    #  ^ 1 error generated.
+    matches = re.search(r"^([^:]+):(\d+):\d+: error: expression result unused", lines[0])
```
brianyu28: Since `expression result unused` could be an `error` or a `warning` in `clang` depending on compiler flags, can you change `error` to `(?:warning|error)` so that we can catch both?

brianyu28:
Thanks for contributing! For organization's sake, can you also move this matcher to right after the `expected expression` one so that the error messages are listed alphabetical?

thctamm:
Done and done

PR #88, opened 2016-09-14
add "division by 0" matcher

brianyu28:
Addressing #72 and `/review` input 2008.


helpers/clang.py
```diff
@@ -114,6 +114,19 @@ def help(lines):
...
+    # foo.c:5:16: error: division by zero is undefined [-Werror,-Wdivision-by-zero]
+    # int x = 28 / 0;
+    #            ^ ~
+    matches = re.search(r"^([^:]+):(\d+):\d+: (?:warning|error): division by zero is undefined", lines[0])
+    if matches:
+        response = [
+            "Looks like you're trying to divide a number by 0 on line {} of `{}`.".format(matches.group(2), matches.group(1)),
+            "Check your math to make sure you don't divide by 0 in your program."
+        ]
+        return (2, response) if len(lines) >= 2 else (1, response)
```
thctamm: If you're only matching on the first line, then why ever return (2, response)?
just return (1, response) seems more appropriate.

thctamm:
Looks good to me

helpers/clang.py
```diff
@@ -114,6 +114,19 @@ def help(lines):
 
     # $ clang foo.c
     # /tmp/foo-1ce1b9.o: In function `main':
+    # foo.c:5:16: error: division by zero is undefined [-Werror,-Wdivision-by-zero]
+    # int x = 28 / 0;
+    #            ^ ~
+    matches = re.search(r"^([^:]+):(\d+):\d+: (?:warning|error): division by zero is undefined", lines[0])
```
dmalan: @brianyu28 is it time for us to start factoring out these regexes somehow, since so many of our `clang` matchers start with this incantation? Perhaps an unexported helper function in `clang.py`? Even when some messages can only be errors, not warnings or errors, seems reasonable to look for `(?:warning|error)`. 

Perhaps the helper could allow us to do, e.g.:

`matches = self.match("division by zero is undefined")`

where `matches` is structured as it always is, but the argument to `match` can also contain matching parens that get used in the regex that `match` uses underneath the hood?

brianyu28: Sounds like a good idea to me. Would probably still need to be `self.match("division by zero is undefined", lines[0])` in order to perform the match, but I'll start working on that.

PR #96, opened 2016-09-15
Match errors with "match" instead of "re.search"

brianyu28:
Per our discussion earlier, I've written `match(expression, line)` in `clang.py`, which takes an error regular expression and a line to search on. `match` will by default search for a `clang` error line with that particular error message.

The `match_warning` optional argument will allows the matcher to match an `error` or a `warning`, the `fatal_error` optional argument specifies to search for `fatal error` instead of just `error`, and the `raw` optional argument is for when we're trying to match something that doesn't follow `clang`'s normal error output (like the linker error we match).

I went ahead and changed our existing matchers to use this format.

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -538,9 +538,11 @@ def help(lines):
     # foo.c:1:1: error: unknown type name 'include'
     # include <stdio.h>
     # ^
-    matches = re.search(r"unknown type name 'include'", lines[0])
+    matches = match(r"unknown type name 'include'", lines[0])
     if matches:
-        response = ["Try including header files via `#include` rather than just `include`."]
+        response = [
+            "Try including header files via `#include` rather than just `include` on line {} of `{}`.".format(matches.group(2), matches.group(1))
+        ]
         return (2, response) if len(lines) >= 2 else (1, response)
 
     # $ clang foo.c
@@ -548,9 +550,11 @@ def help(lines):
     # foo.c:6:9: error: unused variable 'x' [-Werror,-Wunused-variable]
     #     int x = 28;
     #         ^
-    matches = re.search(r"unused variable '([^']+)'", lines[0])
+    matches = match(r"unused variable '([^']+)'", lines[0])
     if matches:
-        response = ["It seems that the variable `{}` is never in your program. Try either removing it altogether or using it.".format(matches.group(1))]
+        response = [
+            "It seems that the variable `{}` (delcared on line {} of `{}`) is never used in your program. Try either removing it altogether or using it.".format(matches.group(3), matches.group(2), matches.group(1))
+        ]
         return (1, response)
 
     # $ clang foo.c
@@ -558,7 +562,7 @@ def help(lines):
     # foo.c:6:20: error: variable 'x' is uninitialized when used here [-Werror,-Wuninitialized]
     #     printf("%d\n", x);
     #                    ^
-    matches = re.search(r"^([^:]+):(\d+):\d+: (?:warning|error): variable '(.*)' is uninitialized when used here", lines[0])
+    matches = match(r"variable '(.*)' is uninitialized when used here", lines[0], match_warning=True)
     if matches:
         response = [
             "It looks like you're trying to use the variable `{}` on line {} of `{}`.".format(matches.group(3), matches.group(2), matches.group(1)),
@@ -568,3 +572,10 @@ def help(lines):
         if len(lines) >= 2 and re.search(matches.group(3), lines[1]):
             return (2, response)
         return (1, response)
+
+def match(expression, line, match_warning=False, fatal_error=False):
+    error = "error" if not fatal_error else "fatal error"
+    if match_warning:
+        error = "(?:warning|" + error + ")"
+    query = r"^([^:]+):(\d+):\d+: " + error + ": " + expression
+    return re.search(query, line)
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -63,7 +63,7 @@ def help(lines):
     # $ clang foo.c
     # /tmp/foo-1ce1b9.o: In function `main':
     # foo.c:7:1: error: control may reach end of non-void function [-Werror,-Wreturn-type]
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -254,8 +254,8 @@ def help(lines):
     matches = match(r"expression result unused", lines[0], match_warning=True)
     if matches:
         response = [
-            "On line {} of `{}` you are doing an operation, but not saving the result.".format(matches.group(2), matches.group(1)),
-            "Try storing the result in a variable"
+            "On line {} of `{}` you are performing an operation, but not saving the result.".format(matches.group(2), matches.group(1)),
+            "Try storing the result in a variable."
```
dmalan: "Did you mean to print or store the result in a variable?" As written, seems to be too likely to be incorrect advice.

dmalan: "Did you mean to print or store the result in a variable?" As written, seems to be too likely to be incorrect advice.

helpers/clang.py
```diff
@@ -351,14 +351,14 @@ def help(lines):
...
     #    printf("hello, world!");
     #    ^
-    matches = re.search(r"implicitly declaring library function '([^']+)'", lines[0])
+    matches = match(r"implicitly declaring library function '([^']+)'", lines[0])
     if matches:
-        if (matches.group(1) == "printf"):
+        if (matches.group(3) == "printf"):
             response = ["Did you forget to `#include <stdio.h>` (in which `printf` is declared) atop your file?"]
-        elif (matches.group(1) == "malloc"):
+        elif (matches.group(3) == "malloc"):
```
dmalan: Should we change these to "if in array" so as to easily support other functions in `stdio` and `stdlib`, as I think you did yesterday with `crypt`?

helpers/clang.py
```diff
@@ -369,9 +369,11 @@ def help(lines):
     #       [-Werror,-Wint-conversion]
     #    int x = "28";
     #        ^   ~~~~
-    matches = re.search(r"incompatible ([^']+) to ([^']+) conversion", lines[0])
+    matches = match(r"incompatible ([^']+) to ([^']+) conversion", lines[0])
```
dmalan: Why the `'`?

helpers/clang.py
```diff
@@ -382,9 +384,11 @@ def help(lines):
     #    printf("%d\n", "hello!");
     #            ~~     ^~~~~~~~
     #            %s
-    matches = re.search(r"^[^:]+:(\d+):\d+: error: format specifies type '[^:]+' but the argument has type '[^:]+'", lines[0])
+    matches = match(r"format specifies type '[^:]+' but the argument has type '[^:]+'", lines[0], match_warning=True)
     if matches:
-        response = ["Be sure to use the correct format code (%i for integers, %f for floating point values, %s for strings) in your string format statement on line {}.".format(matches.group(1))]
+        response = [
+            "Be sure to use the correct format code (%i for integers, %f for floating point values, %s for strings) in your string format statement on line {} of `{}`.".format(matches.group(2), matches.group(1))
```
dmalan: floating-point

helpers/clang.py
```diff
@@ -573,9 +586,13 @@ def help(lines):
...
 
-def match(expression, line, match_warning=False, fatal_error=False):
-    error = "error" if not fatal_error else "fatal error"
+def match(expression, line, match_warning=False, fatal_error=False, raw=False):
+    error = "error"
     if match_warning:
-        error = "(?:warning|" + error + ")"
-    query = r"^([^:]+):(\d+):\d+: " + error + ": " + expression
+        error = "(?:warning|error)"
+    elif fatal_error:
```
dmalan: Hm, adding a `fatal_error` param doesn't seem all that necessary or compelling? Seems reasonable just to search for, e.g.,

```
(?:warning|(?:fatal )?error)
```

no?

helpers/clang.py
```diff
@@ -573,9 +586,13 @@ def help(lines):
             return (2, response)
         return (1, response)
 
-def match(expression, line, match_warning=False, fatal_error=False):
-    error = "error" if not fatal_error else "fatal error"
+def match(expression, line, match_warning=False, fatal_error=False, raw=False):
```
dmalan: Let's document this for future contributors?

dmalan:
experimenting with new UI... Do field comments before merging.

helpers/clang.py
brianyu28: Definitely, we do this already for `implicit declaration of function`, so doing this for `implicitly declaring library function` too makes a lot of sense. Will update.

helpers/clang.py
brianyu28: Agreed that we probably don't need the param. I changed the default error to `(?:fatal )?error`and left the warning one as `(?:warning|error)` since if it's really a fatal error, it won't be a warning when the flags change.

helpers/clang.py
brianyu28: I think I took this from some of the original error matchers that were in `clang.py` where some of the matching groups were enclosed in `'` symbols. Changed this to `(.*)`.

helpers/clang.py
dmalan: Oh, I should elaborate and ask if we even need the `match_warning` param? Given that `clang`'s output is fairly standardized, feels reasonable to just check for 

```
(?:warning|(?:fatal )?error)
```

anytime we do a (non-`raw`) match, even though some things we know can only be errors? I.e., seems okay if we "look for warnings" with our regex even though we know they'll never be warnings, else we put more burden on ourselves (and contributers) to figure out if something could be a warning or just an error, which doesn't really matter to `help50` anyway?

helpers/clang.py
dmalan: `.+` probably better!

brianyu28:
Pushed a change to address all comments.

dmalan:
:shipit:

brianyu28:
Getting a "Required statuses must pass before merging" message which is preventing me from merging. What are our status checks?

PR #97, opened 2016-09-15
Add hardcoded string to segfault matcher.

brianyu28:


brianyu28:
Hm, still getting a "Required statuses must pass before merging" which isn't allowing me to merge.

dmalan:
Buggy. :( I disabled the feature!

PR #114, opened 2016-09-19
Added 'not a directory' matcher.

brianyu28:
Addressing #113

### DIFF
--- a/helpers/bash.py
+++ b/helpers/bash.py
@@ -14,11 +14,22 @@ def help(lines):
 
     # $ cd foo
     # bash: cd: foo: No such file or directory
-    matches = re.search(r"^bash: (.+): No such file or directory", lines[0])
+    matches = re.search(r"^bash: cd: (.+): No such file or directory", lines[0])
     if matches:
         response = [
             "Are you sure `{}` exists?".format(matches.group(1)),
             "Did you misspell `{}`?".format(matches.group(1))
+        ]
+        return (1, response)
+
+    # $ cd foo
+    # bash: cd: foo: Not a directory
+    matches = re.search(r"^bash: cd: (.+): Not a directory", lines[0])
+    if matches:
+        response = [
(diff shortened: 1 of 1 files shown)


helpers/bash.py
```diff
@@ -14,14 +14,25 @@ def help(lines):
...
             "Did you misspell `{}`?".format(matches.group(1))
         ]
         return (1, response)
 
+    # $ cd foo
+    # bash: cd: foo: Not a directory
+    matches = re.search(r"^bash: cd: (.+): Not a directory", lines[0])
+    if matches:
+        response = [
+            "Looks like you're trying to enter a directory which doesn't exist.",
```
dmalan: s/which/that/.

helpers/bash.py
```diff
@@ -14,14 +14,25 @@ def help(lines):
...
+            "Check to make sure `{}` exists and is a directory (as opposed to a file).".format(matches.group(1)),
+            "Try creating a `{}` directory before entering it.".format(matches.group(1))
```
dmalan: Let's avoid potentially misleading direction. Safer would be, e.g., `"Did you mean to create \`{\`} first?"`

helpers/bash.py
dmalan: `"Looks like you're trying to change directories, but \`{}\` doesn't exist."`

brianyu28: Hm, seems to me that in this case, `{}` does exist, it's just not a directory, so perhaps we need to be more careful with our choice of words? If `{}` didn't exist, students would get a `No such file or directory` as opposed to a `Not a directory` error. The reason I think people are getting a `initials: Not a directory` error is because `initials` is in their `pset2` directory but it's the executable file for `initials.c`, rather than a directory.

brianyu28: Addressed your comments and pushed a potential rewording that might address this.

dmalan:
Ah didn't think of that. Makes sense.

PR #115, opened 2016-09-19
Special-cased `check`, `style`, and `submit`. Made detection of inter…

dmalan:
* Special-cased `check`, `style`, and `submit`.
* Made detection of interpreted languages' file extensions case-insensitive.

### DIFF
--- a/helpers/bash.py
+++ b/helpers/bash.py
@@ -1,15 +1,22 @@
 import re
 def help(lines):
 
-    # $ $ foo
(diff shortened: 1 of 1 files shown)


brianyu28:
Looks good. Also reminds me that I've been meaning to add some `check50` and `style50` matchers more generally for things like `Failed to parse` for `style50` and `E_UNKNOWN error` / `invalid ID` for `check50`.

Issue #139, opened 2016-09-23
second parameter of 'main' (argument array) must be of type 'char **'


brianyu28:
```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    caesar.c  -lcs50 -lm -o caesar
caesar.c:5:5: error: second parameter of 'main' (argument array) must be of type 'char **'
int main(int argc, int argv[])
    ^
1 error generated.
make: *** [caesar] Error 1
```

```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    caesar.c  -lcs50 -lm -o caesar
caesar.c:4:5: error: second parameter of 'main' (argument array) must be of type 'char **'
int main(int argc, char argv[]){
    ^
1 error generated.
make: *** [caesar] Error 1
```

```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    caesar.c  -lcs50 -lm -lcrypt -o caesar
caesar.c:5:5: error: second parameter of 'main' (argument array) must be of type 'char **'
int main(int argc, char* argv)
    ^
1 error generated.
make: *** [caesar] Error 1
```

dmalan:
https://github.com/cs50/help50/commit/dcd951e7e9b77bf455e661a12c9d140cd1282f52

PR #167, opened 2016-09-23
Add check for second parameter of main

bw:
Addresses #139

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -490,8 +490,8 @@ def help(lines):
     matches = match(r"second parameter of 'main' (argument array) must be of type 'char **'", lines[0])
     if matches:
         response = [
-            "Looks like you've changed the definition of main() so that the second parameter is not of type 'char **'.",
-            "C requires main(), if you're providing command line arguments, to have the second parameter of that type."
+            "Looks like your declaration of `main` isn't quite right.",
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -481,6 +481,19 @@ def help(lines):
...
+    # $ make caesar.c
+    # clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    caesar.c  -lcs50 -lm -o caesar
+    # caesar.c:5:5: error: second parameter of 'main' (argument array) must be of type 'char **'
+    # int main(int argc, int argv[])
+    #     ^
+    matches = match(r"second parameter of 'main' (argument array) must be of type 'char **'", lines[0])
+    if matches:
+        response = [
+            "Looks like you've changed the definition of main() so that the second parameter is not of type 'char **'.",
+            "C requires main(), if you're providing command line arguments, to have the second parameter of that type."
```
dmalan: Let's change to

```
Looks like your declaration of `main` isn't quite right. Be sure its second parameter is `string argv[]` or some equivalent!
```

which is a bit more succinct and more consistent with how we present `main`.

bw:
Fixed!

dmalan:
Thanks! Merged in separately. https://github.com/cs50/help50/commit/dcd951e7e9b77bf455e661a12c9d140cd1282f52

Issue #164, opened 2016-09-23
'continue' statement not in loop statement


brianyu28:
```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    test.c  -lcs50 -lm -lcrypt -o test
test.c:51:17: error: 'continue' statement not in loop statement
                continue;
                ^
```

dmalan:
https://github.com/cs50/help50/pull/207

PR #171, opened 2016-09-23
'continue' statement not in loop statement

athenakan:
Issue #164

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -60,10 +60,10 @@ def help(lines):
     # /tmp/foo-1ce1b9.o: In function `main':
     # foo.c:51:17: error: 'continue' statement not in loop statement
     #     continue;
-    matches = match(r"'continue' statement(s)? not in loop statement(s)?", lines[0])
+    matches = match(r"'continue' statement not in loop statement", lines[0])
     if matches:
         response = [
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -58,6 +58,18 @@ def help(lines):
 
     # $ clang foo.c
     # /tmp/foo-1ce1b9.o: In function `main':
+    # foo.c:51:17: error: 'continue' statement not in loop statement
+    #     continue;
+    matches = match(r"'continue' statement(s)? not in loop statement(s)?", lines[0])
```
dmalan: Why the `(s)`?

helpers/clang.py
```diff
@@ -58,6 +58,18 @@ def help(lines):
...
+    if matches:
+        response = [
+            "Looks like you placed the command 'continue' outside of the loop, so there's nothing to continue. Make sure that 'continue' is inside of the loop you want to continue."
```
dmalan: Best not to call `continue` a "command", which doesn't really have meaning. Let's change to:

```
Looks like you're trying to use `continue` somewhere other than inside of a loop. Be sure to use that keyword only inside of a `do/while`, `for`, or `while` loop!
```

dmalan:
Thanks! Looks like we incorporated already! https://github.com/cs50/help50/blob/master/helpers/clang.py#L78

Issue #126, opened 2016-09-23
relational comparison result unused


brianyu28:
```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    mario.c  -lcs50 -lm -o mario
mario.c:12:19: error: relational comparison result unused [-Werror,-Wunused-comparison]
    while (height < 0, height < 23);
           ~~~~~~~^~~
1 error generated.
make: *** [mario] Error 1
```

```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    mario.c  -lcs50 -lm -o mario
mario.c:16:26: error: relational comparison result unused [-Werror,-Wunused-comparison]
    for (h = 0; h > 0; h < 23)
                       ~~^~~~
1 error generated.
make: *** [mario] Error 1
```

```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    initials.c  -lcs50 -lm -o initials
initials.c:16:18: error: relational comparison result unused [-Werror,-Wunused-comparison]
        for(argc > 0; argc < 4; argc++)
            ~~~~~^~~
1 error generated.
make: *** [initials] Error 1
```

```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    vigenere.c  -lcs50 -lm -lcrypt -o vigenere
vigenere.c:41:66: error: relational comparison result unused [-Werror,-Wunused-comparison]
    for (int i = 0, c = 0, n = strlen(key), m = strlen(plain); i < m, c < n; i++, c++)
                                                               ~~^~~
1 error generated.
make: *** [vigenere] Error 1
```

dmalan:
https://github.com/cs50/help50/commit/9b640b41cc0d94bf32888abaa9a4fdf418b62e2b

dmalan:
@brianyu28 I added back `has_caret` for this kind of scenario since calling `caret_extract` but not using the result didn't feel quite right, especially since the symbol here is `<`.

PR #172, opened 2016-09-23
comparison result unused

jzhangg85:


helpers/clang.py
```diff
@@ -4,6 +4,13 @@
 def help(lines):
 
     # $ clang foo.c
+    # foo.c:12:19: error: relational comparison result unused [-Weeror,-Wunused-comparison]
+    matches = match(r"relational comparison result unused", lines[0])
+    if matches: 
+        response = ["Looks like you are trying to make a comparison that has no effect probably due to a syntax error inside of your for loop."]
```
dlloyd09: Is a for loop always the place where this error would be given by clang?

helpers/clang.py
```diff
@@ -473,6 +466,13 @@ def help(lines):
         return (1, response)
 
     # $ clang foo.c
+    # foo.c:12:19: error: relational comparison result unused [-Weeror,-Wunused-comparison]
+    matches = match(r"relational comparison result unused", lines[0])
+    if matches: 
+        response = ["Looks like you are trying to make a comparison that has no effect probably due to a syntax error around the conditional comparison or two conflicting comparisons called due to the lack of an operator such as && or ||."]
```
dmalan: Hm, not sure this advice will be all that helpful to a student; it's pretty complex-sounding and also very guessy. Consider the examples at https://github.com/cs50/help50/issues/126.

Perhaps better would be, e.g.:

```
Looks like you're comparing two values on line {} of `{}` but not using the result?
```

And if `len(lines) >= 3`, we should include `lines[1]` and `lines[2]` in the return value, as with:

```
return (lines[0:3], response);
```

New syntax as of this weekend.

CC @brianyu28

helpers/clang.py
```diff
@@ -473,6 +466,13 @@ def help(lines):
         return (1, response)
 
     # $ clang foo.c
+    # foo.c:12:19: error: relational comparison result unused [-Weeror,-Wunused-comparison]
```
dmalan: Do add more of the example here, as in https://github.com/cs50/help50/issues/126. Else we'll never remember what kind of code this is referring to.

dmalan:
https://github.com/cs50/help50/commit/9b640b41cc0d94bf32888abaa9a4fdf418b62e2b

PR #175, opened 2016-09-23
Forgot condition for loops

Schmill731:
Address 111

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -142,7 +142,7 @@ def help(lines):
     matches = match(r"expected '\(' after '(.*)'", lines[0])
     if matches:
         response = [
-            "Looks like you forget to specify the condition of your `{}` loop! You need to specify under what conditions this loop should run".format(matches.group(3), matches.group(2), matches.group(1))
+            "Looks like you forget to specify the condition of your {} loop! You need to specify under what conditions this loop should run".format(matches.group(3), matches.group(2), matches.group(1))
         ]
         return (3, response)
 
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -133,6 +133,21 @@ def help(lines):
         return (1, response)
 
     # $ clang foo.c
+    # clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    mario.c  -lcs50 -lm -o mario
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -133,6 +133,21 @@ def help(lines):
...
+    # clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    mario.c  -lcs50 -lm -o mario
+    # foo.c:11:5: error: expected '(' after 'do/while'
+    #     {
+    #     ^
+    # 1 error generated.
+    # make: *** [mario] Error 1
+    matches = match(r"expected '\(' after '(.*)'")
+    if matches:
+        response = [
+            "Looks like you forget to specify the condition of your {} loop! You need to specify under what conditions this loop should run".format(matches.group(3), matches.group(2), matches.group(1))
```
dmalan: Hm, will this error definitely signify a loop? (@brianyu28, @glennholloway?) Might be safest to check that `group(3)` is `do/while` or `for` or `while`, and only output "loop" in those cases, having some default language otherwise?

And do surround the {} with backticks, as in other examples.

helpers/clang.py
```diff
@@ -142,7 +142,7 @@ def help(lines):
     matches = match(r"expected '\(' after '(.*)'", lines[0])
     if matches:
         response = [
-            "Looks like you forget to specify the condition of your {} loop! You need to specify under what conditions this loop should run".format(matches.group(3), matches.group(2), matches.group(1))
+            "Looks like you forget to specify the condition of your `{}` loop! You need to specify under what conditions this loop should run".format(matches.group(3), matches.group(2), matches.group(1))
```
dmalan: @Schmill731 let's indeed check that `group(3)` is `do/while` or `for` or `while`, since this error message from `clang` is generic enough that it's probably used in other contexts. But do just whip up 3 simple samples to make sure that this is indeed the error message that's triggered when the conditions are missing.

Schmill731: @dmalan Done! I verified it by making a program on the IDE that had 3 loops, each of which had no condition. File attached.
[test_help50.zip](https://github.com/cs50/help50/files/491633/test_help50.zip)

PR #176, opened 2016-09-23
Fixed #140

### LINKED ISSUE #140: assigning from incompatible type
```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    greedy.c  -lcs50 -lm -o greedy
greedy.c:8:11: error: assigning to 'float' from incompatible type 'float (void)'
        f = get_float;
          ^ ~~~~~~~~~
1 error generated.
make: *** [greedy] Error 1
```

tlively:


helpers/clang.py
```diff
@@ -37,6 +36,15 @@ def help(lines):
         return (1, response)
 
     # $ clang foo.c
+    # foo.c:row:col: error: assigning to 'type' from incompatible type 'type (argtypes)'
```
dmalan: @tlively, generalizing `row` and `col` is okay, but let's change `type (argtypes)` back to the actual error in question. Less obvious what the sample output actually is for future testing otherwise. E.g., I just had to refer back to the issue, https://github.com/cs50/help50/issues/140, to see what the original output was.

helpers/clang.py
```diff
@@ -37,6 +36,15 @@ def help(lines):
...
+    #         f = get_float;
+    #           ^ ~~~~~~~~~
+    matches = match(r"assigning to '(.+)' from incompatible type '\3 \(.+\)'", lines[0])
```
dmalan: This seems a bit fragile, since we'd no longer be able to handle a scenario like

```
int i = get_float
```

or such.

helpers/clang.py
```diff
@@ -37,6 +36,15 @@ def help(lines):
...
+    if matches:
+        func = tilde_extract(lines[1:]) if len(lines) >= 3 else None
+        return (1, ["It looks like you're trying to call the function {}. Did you forget parentheses?".format(func)])
```
dmalan: Let's put backticks around the `{}`.

helpers/clang.py
tlively: Done.

helpers/clang.py
tlively: Done.

helpers/clang.py
```diff
@@ -37,6 +36,15 @@ def help(lines):
         return (1, response)
 
     # $ clang foo.c
+    # foo.c:row:col: error: assigning to 'float' from incompatible type 'float (void)'
+    #         f = get_float;
+    #           ^ ~~~~~~~~~
+    matches = match(r"assigning to '(.+)' from incompatible type '.+ \(.+\)'", lines[0])
+    if matches:
+        func = tilde_extract(lines[1:]) if len(lines) >= 3 else None
+        return (1, ["It looks like you're trying to call the function `{}`. Did you forget parentheses?".format(func)])
```
dmalan: `func` can be `None` but you're still passing it to `format`?

helpers/clang.py
tlively: Done.

helpers/clang.py
tlively: Done.

helpers/clang.py
```diff
@@ -41,8 +41,8 @@ def help(lines):
     #           ^ ~~~~~~~~~
     matches = match(r"assigning to '(.+)' from incompatible type '.+ \(.+\)'", lines[0])
     if matches:
-        func = tilde_extract(lines[1:]) if len(lines) >= 3 else None
-        return (1, ["It looks like you're trying to call the function `{}`. Did you forget parentheses?".format(func)])
+        func = "`{}`".format(tilde_extract(lines[1:])) if len(lines) >= 3 else "a function"
+        return (1, ["It looks like you're trying to call {}. Did you forget parentheses?".format(func)])
```
dmalan: See other matchers for our convention for mentioning filename/line. "... on line {} of `{}`, but you seem to have forgotten parentheses after the function's name?"

Also, since Friday, we changed the convention for return values: just change the `1` to `lines[0:1]`.

helpers/clang.py
```diff
@@ -42,7 +42,8 @@ def help(lines):
     matches = match(r"assigning to '(.+)' from incompatible type '.+ \(.+\)'", lines[0])
     if matches:
         func = "`{}`".format(tilde_extract(lines[1:])) if len(lines) >= 3 else "a function"
```
dmalan: Still not safe, since `tilde_extract` could return `None`.

tlively: Darn :(

dmalan:
Tweaked here, https://github.com/cs50/help50/commit/572639318eeb9d976b26cbe59309b29e86a446c0.

dmalan:
Thanks!

PR #177, opened 2016-09-23
issue 128 resolve

jzuckerman:


dmalan:
@brianyu28, @glennholloway I'm not sure we can actually parse this one deterministically, given the generic nature of the error message?

brianyu28:
Yeah, this seems like a case where we might want to adopt an approach similar to what we did here:

https://github.com/cs50/help50/blob/master/helpers/clang.py#L306

dmalan:
Thanks, @jzuckerman! Going to take a more conservative approach to be safe! https://github.com/cs50/help50/commit/05f356410221c5dec6dad8fa6b86afa2c21f4c66

PR #180, opened 2016-09-23
Handles unknown type name errors in clang

nhawke:
Added a check (after the existing check for "unknown type 'include'") to handles unknown types.

Suggested action is to ensure custom types are `typedef`'d and that all necessary libraries are `#include`'d

This fixes issue 112


dmalan:
Thanks! Merged in with some tweaks, https://github.com/cs50/help50/commit/383110e51a8a36b26e3a327b7d40678d3e886d92!

PR #185, opened 2016-09-24
seems to work reasonably well and doesn't break too much stuff. Or if…

Xctrunner:
… it does, Brian can fix it, right?


dmalan:
Seems someone did this already!

Issue #159, opened 2016-09-23
void function should not return a value


brianyu28:
```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -o find find.c helpers.c -lcs50 -lm
helpers.c:64:10: error: void function 'sort' should not return a value [-Wreturn-type]
         return 0;
         ^      ~
1 error generated.
make: *** [find] Error 1
```

dmalan:
https://github.com/cs50/help50/commit/2715e55a9ebce50a0a0621fc4c3b0585e7bd22a8

PR #187, opened 2016-09-24
Attempted to add support for #159

panchoruy:
I can't test this, but I think it shouldn't be too far from correct. Please check before merging.

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -588,6 +588,24 @@ def help(lines):
             return (2, response)
         return (1, response)
 
+    # $ clang foo.c
+    # foo.c:6:10: error: void function 'f' should not return a value [-Wreturn-type]
+    #          return 0;
+    #          ^      ~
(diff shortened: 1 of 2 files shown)


helpers/clang.py
```diff
@@ -588,6 +588,24 @@ def help(lines):
             return (2, response)
         return (1, response)
 
+    # $ clang foo.c
+    # foo.c:6:10: error: void function 'f' should not return a value [-Wreturn-type]
+    #          return 0;
+    #          ^      ~
+    matches = match(r"void function '(.*)' should not return a value", lines[0])
+    if matches:
+        return_value = tilde_extract(lines[1:3])
```
dmalan: Careful, what if second and third lines don't exist?

dmalan:
Addressed some issues and merged with https://github.com/cs50/help50/commit/2715e55a9ebce50a0a0621fc4c3b0585e7bd22a8, thanks!

Issue #131, opened 2016-09-23
ignoring return value of function declared with const attribute


brianyu28:
```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    round.c  -lcs50 -lm -o round
round.c:17:5: error: ignoring return value of function declared with const attribute [-Werror,-Wunused-value]
    round(cents);
    ^~~~~ ~~~~~
1 error generated.
make: *** [round] Error 1
```

```
clang -ggdb3 -O0 -std=c11 -Wall -Werror -Wshadow    caesar.c  -lcs50 -lm -lcrypt -o caesar
caesar.c:9:5: error: ignoring return value of function declared with pure attribute [-Werror,-Wunused-value]
    atoi(argv[1]);
    ^~~~ ~~~~~~~
1 error generated.
make: *** [caesar] Error 1
```

dmalan:
https://github.com/cs50/help50/commit/ebcb79cd8198904e9b21e6dfd4493e395ab36e16

PR #188, opened 2016-09-24
Issue 131 ignored return handling

raphaelrk:
lightly tested on localhost
Fix for #131

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -333,12 +333,12 @@ def help(lines):
     matches = match(r"ignoring return value of function declared with (\w+) attribute", lines[0])
     if matches:
         response = [
-            "You seem to have an error in `{}` on line {}.".format(matches.group(1), matches.group(2))
+            "You seem to be calling a function on line {} of `{}` but aren't using its return value.".format(matches.group(1), matches.group(2))
         ]
         if len(lines) > 1:
-            response.append("Did you forget to assign the result of '" + lines[1] + "' to a variable?")
+            response.append("Did you mean to assign the result of '" + lines[1] + "' to a variable?")
         else:
-            response.append("Did you forget to assign the result of that line to a variable?")
+            response.append("Did you mean to assign it to a variable?")
         return (1, response)
 
     # $ clang foo.c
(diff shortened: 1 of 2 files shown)


helpers/clang.py
```diff
@@ -326,6 +326,22 @@ def help(lines):
...
+    # round(cents);
+    # ^~~~~ ~~~~~
+    #
+    matches = match(r"ignoring return value of function declared with (\w+) attribute", lines[0])
+    if matches:
+        response = [
+            "You seem to have an error in `{}` on line {}.".format(matches.group(1), matches.group(2))
+        ]
+        if len(lines) > 1:
+            response.append("Did you forget to assign the result of '" + lines[1] + "' to a variable?")
```
dmalan: Doesn't seem safe to print whole line. Do use @brianyu28's `tilde_extract` method as needed. 

Let's change "forget" to "mean".

helpers/clang.py
```diff
@@ -326,6 +326,22 @@ def help(lines):
...
 
     # $ clang foo.c
+    # round.c:17:5: error: ignoring return value of function declared with const attribute [-Werror,-Wunused-value]
+    # round(cents);
+    # ^~~~~ ~~~~~
+    #
+    matches = match(r"ignoring return value of function declared with (\w+) attribute", lines[0])
+    if matches:
+        response = [
+            "You seem to have an error in `{}` on line {}.".format(matches.group(1), matches.group(2))
```
dmalan: Instead of this, perhaps more descriptively:

```
You seem to be calling a function on line {} of `{}` but aren't using its return value.
```

helpers/clang.py
```diff
@@ -326,6 +326,22 @@ def help(lines):
...
+        ]
+        if len(lines) > 1:
+            response.append("Did you forget to assign the result of '" + lines[1] + "' to a variable?")
+        else:
+            response.append("Did you forget to assign the result of that line to a variable?")
```
dmalan: "Did you mean to assign it to a variable?"

raphaelrk:
@dmalan added what you asked for, had to make a separate tilde extract for functions as the current one just returned "ound" instead of "round(cents)"

dmalan:
Thanks! Can use `var_extract` for functions, actually. :) Will consider how to rename! https://github.com/cs50/help50/commit/ebcb79cd8198904e9b21e6dfd4493e395ab36e16

@brianyu28

PR #196, opened 2016-09-25
adds 'before' text to ANSI output, highlighting it in yellow

dmalan:


dmalan:
@brianyu28, @glennholloway, am preserving colorized output (from, e.g., `clang`) now thanks to `script`, https://github.com/cs50/help50/pull/196/commits/adb516536e1f380d914431c7451434939dd36941#diff-5306b6a56f5cfe829bb55cd7d8c6eb77R23.

dmalan:
screen shot 2016-09-25 at 2 12 00 pm

screen shot 2016-09-25 at 2 12 09 pm

brianyu28:
One thing I'm noticing is that for the `Sorry, not sure how to help` messages, `before` is just all of the lines: so the user ends up seeing all of their error messages duplicated, followed by a `not sure how to help` message. Would it be better for `before` to be empty in this case, such that the student only sees each message once if we're not trying to draw their attention to any lines in particular?

Aside from that, looks good.

dmalan:
Hmm, how did you induce that behavior? `unhelpful.ans` only outputs the apology, not the lines again.

https://github.com/cs50/help50/pull/196/files#diff-7252d8c80935263ce1723cc5c4a274fd

brianyu28:
Ah nevermind, looks like I was testing this branch by making requests to `help.cs50.net`, which still currently includes `before` in its `.ans` output.

Looks good!

PR #198, opened 2016-09-25
"before" and valgrind.py

brianyu28:
Changes here:

- Revert to helper functions responding with two lists of strings, rather than a number and a list of strings.
- Added a `valgrind.py` file for valgrind matchers. Starting off by just matching "definitely lost" and "use of uninitialized value." Also created an `issue_locate` function with a tentative solution for finding the source of a memory bug: currently it just looks for the first line at which the filename changes. For example, feeding the following into `issue_locate`:

```
==26==    at 0x4E8099B: _itoa_word (_itoa.c:179)
==26==    by 0x4E84636: vfprintf (vfprintf.c:1660)
==26==    by 0x4E8B3D8: printf (printf.c:33)
==26==    by 0x40054C: main (foo.c:6)
```

will return the tuple ("main", "foo.c", 6), because it guesses that we're not interested in what's in `printf.c` since that's in a different file. For multi-file programs this has the potential to be problematic, so at some point it would be good to figure out a better solution to this (I thought this was cleaner than hard-coding "printf" and other library functions into `valgrind.py`).

### DIFF
--- a/application.py
+++ b/application.py
@@ -53,9 +53,12 @@ def index():
                 # helpful response
                 if help:
                     before, after = help
+                    if (len(before) < len(lines[i:])):
+                        before.append("...")
+                    before = "\n".join(before)
                     after = " ".join(after)
                     model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), after)
-                    return render_template("helpful." + format, before="\n".join(before), after=after)
+                    return render_template("helpful." + format, before=before, after=after)
 
         # unhelpful response
         model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), None)
--- a/application.py
+++ b/application.py
@@ -52,10 +52,10 @@ def index():
 
                 # helpful response
                 if help:
-                    n, response = help
-                    response = " ".join(response)
-                    model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), response)
-                    return render_template("helpful." + format, before="\n".join(lines[:i+n]), after=response)
+                    before, after = help
+                    after = " ".join(after)
+                    model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), after)
+                    return render_template("helpful." + format, before="\n".join(before), after=after)
 
         # unhelpful response
         model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), None)
--- a/helpers/valgrind.py
+++ b/helpers/valgrind.py
@@ -1,4 +1,5 @@
 import re
+from collections import namedtuple
 
 def help(lines):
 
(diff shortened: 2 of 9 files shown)


application.py
```diff
@@ -52,10 +52,10 @@ def index():
...
                 # helpful response
                 if help:
-                    n, response = help
-                    response = " ".join(response)
-                    model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), response)
-                    return render_template("helpful." + format, before="\n".join(lines[:i+n]), after=response)
+                    before, after = help
+                    after = " ".join(after)
+                    model.log(request.form.get("cmd"), request.form.get("username"), request.form.get("script"), after)
+                    return render_template("helpful." + format, before="\n".join(before), after=after)
```
dmalan: For consistency with `before`, should probably be

```
after=" ".join(after)
```

?

dmalan: And shall we do

```
if before is shorter than lines[i:], append "..."
```

?

helpers/valgrind.py
```diff
@@ -0,0 +1,12 @@
+import re
+
+def help(lines):
+
+    # definitely lost: 4 bytes in 1 blocks
+    matches = re.search(r"definitely lost: (\d+) bytes in (\d+) blocks", lines[0])
+    if matches:
+        response = [
+            "Looks like your program leaked {} bytes worth of memory.".format(matches.group(1)),
```
dmalan: -worth

helpers/valgrind.py
```diff
@@ -10,3 +11,45 @@ def help(lines):
             "Did you forget to `free` memory that you allocated using `malloc`?"
         ]
         return (lines[0:1], response)
+
+    # use of uninitialized value of size 8
+    matches = re.search(r"Use of uninitialised value of size (\d+)", lines[0])
+    if matches:
+        response = [
+            "Looks like you're trying to access the contents of a variable which you may not have given a value.",
```
dmalan: Feels a bit hard to parse. Perhaps "Looks like you're trying to access a variable that you might not have assigned a value?"

helpers/valgrind.py
```diff
@@ -10,3 +11,45 @@ def help(lines):
...
+            "Check to make sure that you initialize your variables (by setting their initial values) before using them."
```
dmalan: Sounds redundant.

helpers/valgrind.py
```diff
@@ -10,3 +11,45 @@ def help(lines):
...
+        ]
+        issue = issue_locate(lines[1:])
+        if issue:
+            response.append("In particular, pay attention to line {} of `{}`.".format(issue.line, issue.file))
```
dmalan: Nice.

dmalan:
Really great overall. Thanks so much for changing so many tuples!

dmalan:
:shipit:

Issue #203, opened 2016-09-26
use has_caret


dmalan:
@brianyu28, in quite a few places, we seem to be assuming that `lines[2]` will have [caret diagnostics](http://clang.llvm.org/diagnostics.html), but that's not necessarily true if `clang` is run with other flags, in which case our code might 500 because of `None` return values. We should probably adopt a paradigm of

```python
if len(lines) > 3 and has_caret(lines[2]):
```

or even combine that? I've added `has_caret` to https://github.com/cs50/help50/blob/master/tools.py.

brianyu28:
`var_extract` (which probably should be named `caret_extract`) and `tilde_extract` are both set to return `None` if they can't extract for some reason (such as if there is no caret). Looking back, I realize that there are some cases where I forgot to check for `None`, I can go back and fix those.

dmalan:
Sounds good. Should those perhaps be at the bottom of `clang.py` instead, where `match` is, since they're `clang`-specific?

PR #204, opened 2016-09-26
caret_extract

brianyu28:
Per discussion in #203


helpers/clang.py
```diff
@@ -386,8 +385,8 @@ def help(lines):
     # ^~~~~ ~~~~~
     matches = match(r"ignoring return value of function declared with (.+) attribute", lines[0])
     if matches:
-        if len(lines) >= 3 and has_caret(lines[2]):
```
dmalan: RIP `has_caret`.

helpers/clang.py
```diff
@@ -776,3 +775,51 @@ def match(expression, line, raw=False):
...
+def char_extract(lines):
+    if len(lines) < 2 or not re.search(r"\^", lines[1]):
+        return
+    index = lines[1].index("^")
+    if len(lines[0]) < index + 1:
+        return
+    return lines[0][lines[1].index("^")]
+
+# returns True if line has caret diagnostics (and possibly range highlighting)
+def has_caret(line):
```
dmalan: Needed anymore?

helpers/clang.py
```diff
@@ -776,3 +775,51 @@ def match(expression, line, raw=False):
...
+    return re.search(r"^[ ~]*\^[ ~]*$", line) is not None
+
+# extracts all characters above the first sequence of ~
+def tilde_extract(lines):
```
dmalan: Needed?

helpers/clang.py
```diff
@@ -776,3 +775,51 @@ def match(expression, line, raw=False):
...
+    while len(lines[1]) > start + length and lines[1][start + length] == "~":
+        length += 1
+    if len(lines[0]) < start + length:
+        return
+    return lines[0][start:start+length]
+
+# extract the name of a variable above the ^
+# by default, assumes that ^ is under the first character of the variable
+# if left_aligned is set to False, ^ is under the next character after the variable
+def caret_extract(lines, left_aligned=True):
```
dmalan: Should we add a named parameter that limits return value to a single char, to replace `char_extract`?

helpers/clang.py
```diff
@@ -776,3 +775,51 @@ def match(expression, line, raw=False):
     if raw:
         query = expression
     return re.search(query, line)
+
+# extracts a single character above the ^
+def char_extract(lines):
```
dmalan: Needed?

helpers/clang.py
```diff
@@ -776,3 +775,51 @@ def match(expression, line, raw=False):
...
+        return
+    return lines[0][start:start+length]
+
+# extract the name of a variable above the ^
+# by default, assumes that ^ is under the first character of the variable
+# if left_aligned is set to False, ^ is under the next character after the variable
+def caret_extract(lines, left_aligned=True):
+    if len(lines) < 2 or not re.search(r"\^", lines[1]):
+        return
+    permissibles = string.ascii_letters + string.digits + '_'
```
dmalan: Can we not rely on range highlighting (tildes) in order to extract symbols, rather than iterating over valid chars?

helpers/clang.py
brianyu28: Probably a good idea, will do.

helpers/clang.py
brianyu28: Unfortunately not always, `clang` seems to be pretty inconsistent about this. Sometimes tildes are present, as per:

```
$ clang mario.c
foo.c:3:1: error: 'main' must return 'int'
void main(void)
^~~~
int
```

but in other cases we'll get:

```
$ clang foo.c
foo.c:3:1: error: type specifier missing, defaults to 'int' [-Werror,-Wimplicit-int]
square (int x) {
^
```

Given that, feels safest to extract the token above the caret by iterating over valid characters.

dmalan: Ah, interesting, makes sense.

brianyu28:
Okay, consolidated the helper functions into `tilde_extract` and `caret_extract`. Both return `None` if unable to extract, and so if matchers use either function, they should check to see if the return value was `None` and adjust the `response` appropriately.

helpers/clang.py
```diff
@@ -776,3 +776,41 @@ def match(expression, line, raw=False):
...
+            var += lines[0][index]
+            index += 1
+    elif len(lines[0]) > index + 1:
+        index -= 1
+        while index >= 0 and lines[0][index] in permissibles:
+            var = lines[0][index] + var
+            index -= 1
+
+    if len(var) > 0:
+        return var
```
dmalan: Could this not be shortened significantly simply by using a regex, whereby you match

```
r"^[A-Za-z0-9_]+"
```

against `lines[0][index:]`?

brianyu28: Yeah, that's much nicer. Fixed!

dmalan:
love the regex. :shipit:

PR #207, opened 2016-09-26
Fixed issue 164

tlively:
#164


helpers/clang.py
```diff
@@ -75,6 +75,18 @@ def help(lines):
         return (lines[0:1], response)
 
     # $ clang foo.c
+    # test.c:51:17: error: 'continue' statement not in loop statement
+    #                 continue;
+    #                 ^
+    matches = match(r"'continue' statement not in loop statement", lines[0])
+    if matches:
+        response = [
+            "Looks like you're trying to use the `continue` keyword on line {} of `{}`, which isn't inside a loop. `continue` jumps to the beginning of the enclosing loop, so you can't use it outside a loop."
```
dmalan: Let's shorten to:

```
Looks like you're trying to use `continue` on line {} of `{}`, which isn't inside of a loop, but that keyword can only be used inside of a loop.
```

And for stylistic consistency, let's keep the `.format(...` on the same line.

dmalan:
Thanks!

PR #209, opened 2016-09-27
Update clang match function

brianyu28:
Addressing #179 

In the old version of `match`:

- `matches.group(1)` was the filename
- `matches.group(2)` was the line number
- `matches.group(3)`, `matches.group(4)`, `matches.group(5)`, etc. were the first, second, and third matches requested by the argument to the `match` function.

In this version:

- `matches.file` is the filename
- `matches.line` is the line number
- `matches.group[0]`, `matches.group[1]`, `matches.group[2]`, etc. are the first, second, and third matches.

Given that `matches.group` is now a list (instead of a function), I opted for zero-indexing it, but if you think the semantics of `matches.group[1]` for the first group is nicer, that's a fairly simple change.

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -8,7 +8,7 @@ def help(lines):
     # foo.c:13:25: error: adding 'int' to a string does not append to the string [-Werror,-Wstring-plus-int]
     matches = match(r"adding '(.+)' to a string does not append to the string", lines[0])
     if matches:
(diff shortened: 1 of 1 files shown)


dmalan:
This is great. Thanks so much. Zero-indexing sounds reasonable since it's now a list. Let's see if it causes any confusion. Can always switch!

:shipit:

PR #212, opened 2016-09-28
Resizes /review table to fit viewport

coltonoscopy:
Makes necessary CSS and HTML changes to ensure /review table fits to the viewport 100% (with slight padding). Table resizes dynamically when window changes in size.

### DIFF
--- a/templates/layout.html
+++ b/templates/layout.html
@@ -36,15 +36,18 @@
             table, th, td {
                 border: 1px solid black;
                 padding: 5px;
+                white-space: pre;
             }
             
             table {
                 width: 100%;
                 table-layout: fixed;
+                margin: 0;
             }
             
             td {
                 text-overflow: hidden;   
+                overflow: hidden;
             }
 
             label {
@@ -58,6 +61,10 @@
                 display: block;
                 top: 50%;
                 position: relative;
+            }
+            
+            .container {
+                width: 100%;   
             }
 
         </style>
--- a/templates/layout.html
+++ b/templates/layout.html
@@ -21,7 +21,7 @@
         <script src="https://cdn.jsdelivr.net/jquery.autosize/3.0.17/autosize.min.js"></script>
 
         <style>
-            
+
             body {
                 font-family: Hack, monospace;
                 margin-top: 20px;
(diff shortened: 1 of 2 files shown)


templates/review.html
```diff
@@ -5,21 +5,21 @@
...
-                <th>Reviewed</th>
-                <th>id</th>
-                <th>cmd</th>
-                <th>timestamp</th>
-                <th>script</th>
+                <th width="7%">Reviewed</th>
+                <th width="5%">id</th>
+                <th width="15%">cmd</th>
+                <th width="10%">timestamp</th>
+                <th width="63%">script</th>
```
dmalan: Deprecated in HTML5, I think. Better to use CSS selectors and have the template include the CSS file somehow!

templates/layout.html
```diff
@@ -37,6 +38,17 @@
                 padding: 5px;
                 white-space: pre;
             }
+            
+            table {
+                width: 100%;
+                table-layout: fixed;
+                margin: 0;
```
dmalan: Let's alphabetize these and other properties for style's sake.

coltonoscopy:
All set!

templates/review.html
```diff
@@ -1,6 +1,11 @@
 {% extends "layout.html" %}
 {% block body %}
 {% if inputs | length > 0 %}
+<style>
+    .container {
+        width: 100%;
+    }
+</style>
```
dmalan: Pretty sure HTML5 requires that this be in `head`. See https://github.com/cs50/help50/commit/688d3e11480bfcec0e35b0c76b33c6a155f6911c for a placeholder for you, whereby templating will put it where you want.

templates/review.html
```diff
@@ -1,6 +1,11 @@
 {% extends "layout.html" %}
 {% block body %}
 {% if inputs | length > 0 %}
+<style>
+    .container {
+        width: 100%;
```
dmalan: Is this necessary? Does `.container` not default to 100% in Bootstrap? Just curious.

templates/layout.html
```diff
@@ -21,34 +21,66 @@
...
+                margin: auto;
                 position: relative;
+                top: 50%;
+            }
+
+            .th_reviewed {
+                width: 7%;
+            }
+
+            .th_id {
```
dmalan: I'd remove the `th_` prefix. Blurring the lines between metadata and markup. Plus, you're using these inside of `td`s too.

templates/layout.html
```diff
@@ -21,34 +21,66 @@
...
+                width: 5%;
```
dmalan: Feels unnecessary to prescribe such specific percentages. Why not let Reviewed, id, and timestamp autosize based on their width, and then use `max-width` or such for `cmd` and `script`, which are the problematic ones?

templates/review.html
coltonoscopy: It doesn't; is something like 70 or 75%!

templates/review.html
coltonoscopy: Relocated to style tags!

templates/layout.html
coltonoscopy: Removed!

templates/layout.html
coltonoscopy: max-width doesn't fix (and would require either hardcoded pixels or percentages anyway). I'm not great at CSS, so I asked Jordan, who said we should be using Flex anyway for this sort of thing, but feels like too much to do a Flex refactor of the table. Do we need to worry about this detail, given this page will only be seen by very few folks anyway and the layout seems to work perfectly well as is? If you do think so, do you have other ideas in mind as to how to achieve? If not, I can ask Luke tomorrow what he thinks would be best.

dmalan:
Tidied a bit! https://github.com/cs50/help50/commit/d35510c65922003bcffdcb7ebb91b209bdad0342

coltonoscopy:
Looks great!

PR #214, opened 2016-09-28
Fix permission denied regex

brianyu28:
Addressing the "unable to help" message that came from 

```
bash: ./foo: Permission denied
```

Turns out this isn't a new issue at all, it's just a problem with the regex for the error that's been there since the beginning. The old regex would only match "permission denied" errors for `./foo.c`, `./foo.php`, etc. (with an extension), and wouldn't match `./foo`.

I fixed the regex, and now it should work for both `./foo` and `./foo.c`

### DIFF
--- a/helpers/bash.py
+++ b/helpers/bash.py
@@ -41,7 +41,7 @@ def help(lines):
 
     # $ ./foo
     # bash: ./foo: Permission denied
-    matches = re.search(r"^bash: .*?(([^/]+)\.([^/.]+)): Permission denied", lines[0])
+    matches = re.search(r"^bash: .*?(([^/]+?)\.?([^/.]*)): Permission denied", lines[0])
     if matches:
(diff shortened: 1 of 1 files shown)


helpers/bash.py
```diff
@@ -41,7 +41,7 @@ def help(lines):
 
     # $ ./foo
     # bash: ./foo: Permission denied
-    matches = re.search(r"^bash: .*?(([^/]+)\.([^/.]+)): Permission denied", lines[0])
+    matches = re.search(r"^bash: .*?(([^/]+?)\.?([^/.]*)): Permission denied", lines[0])
```
dmalan: Want to add "Does `{}` definitely exist?" to `response`?

dmalan: nvm, I can do it!

dmalan:
Surprised we didn't notice this one already! Thanks!

PR #215, opened 2016-09-28
Unescape HTML characters

brianyu28:
While testing a separate issue, noticed that characters like `'` in `help50` were being turned into `&#39;` in help50's output, as per the below:

screen shot 2016-09-28 at 1 27 55 pm

Went into the `filter` that we use for `helpful.ans` and updated it to unescape the HTML characters. Before and after are shown below:

screen shot 2016-09-28 at 1 48 41 pm

### DIFF
--- a/application.py
+++ b/application.py
@@ -3,6 +3,7 @@ from flask_migrate import Migrate
 from flask_sqlalchemy import SQLAlchemy
 import flask_migrate
 import helpers
+import html as HTML
 import manage
 import model
 import os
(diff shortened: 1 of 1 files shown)


application.py
```diff
@@ -102,7 +103,7 @@ def bad_request(e):
 # ANSI filter
 @app.template_filter("ans")
 def ans(value):
-    return re.sub(r"`([^`]+)`", r"\033[1m\1\033[22m", value)
+    return re.sub(r"`([^`]+)`", r"\033[1m\1\033[22m", HTML.unescape(value))
```
dmalan: So I think the simpler solution would be just this?

https://github.com/cs50/help50/commit/81a7a4cf7f51081c07db94c8d4fabbe9f8764194

I suspect I copied/pasted the HTML template a while back and left in too much filtration.

brianyu28: Ah okay, looks good. Thanks.

Issue #104, opened 2016-09-18
make: No targets specified and no makefile found.  Stop.


brianyu28:
```
make: *** No targets specified and no makefile found.  Stop.
```

PR #216, opened 2016-09-28
added help for no target specified and no makefile found issue #104

Bujo0:


helpers/make.py
```diff
@@ -30,6 +30,15 @@ def help(lines):
             "If using a Makefile, are you sure you have a target called `{}`?".format(matches.group(1))
         ]
         return (1, response)
+
+    # $ make
+    # make: *** No targets specified and no makefile found.  Stop.
+    matches = re.search(r"^make: \*\*\* No targets specified and no makefile found. Stop", lines[0])
```
dmalan: Looks like there should be two spaces before `Stop`?

helpers/make.py
```diff
@@ -30,6 +30,15 @@ def help(lines):
...
+    if matches:
+        response = [
+            "You have not created a makefile."
```
dmalan: Better if advice isn't so definitive, lest we be wrong. And better to ask questions than outright state conclusions.

dmalan:
Tweaked and incorporated here, thanks! https://github.com/cs50/help50/commit/220efe1448e250ec30501c40f2b93b0a7ea95831

PR #217, opened 2016-09-29
Sorts /review entries based on similar "script" fields (descending frequency)

coltonoscopy:
Utilizes the fuzzywuzzy Python library for fuzzy string matching. Summary of the sorting algorithm:

`let C = map of similar-matching string lists`
`for every record in the database:`
&nbsp;&nbsp;&nbsp;`let L = most relevant line in record script field to fuzzy match`
&nbsp;&nbsp;&nbsp;`for every collection of lists in C:`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`for every ratio threshold from 90 to 50 counting down by 10:`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`let R = [] to hold fuzzy-matching ratios`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`for every entry in collection:`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`add ratio of fuzzy match between L and entry to R`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`if average of R > ratio threshold:`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`add L to collection`

Collection of lists is then sorted in descending order of collection size and flattened before being returned to the template. This whole thing runs every time the /review route is reached and does not cache results, so it needs to be recomputed each time, which is highly inefficient.

Requires

`pip install fuzzywuzzy`

to work, and

`pip install python-Levenshtein`

to work faster.

The process is slow (~60 seconds) by the nature of string-matching hundreds of entries against each other to sort them into buckets each time we query the database. This is many times faster than it was before (> 10 minutes), when the algorithm was to compare full script fields against full script fields, but that is less helpful because of the redundancy of "clang" lines and the last two lines of each script more generally, which skew the data for fuzzy pattern matching, given they are almost always very similar, hence the refining to just lines[1] or lines[0]. 

Curious to hear what might be best in terms of optimizing this so it takes less time than a minute. At first, I thought it might be possible to cache results some way, but it still needs to get a fuzzy ratio for every string in the database per record no matter what, so I'm not sure that would help in any way.

### DIFF
--- a/model.py
+++ b/model.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 from flask_sqlalchemy import SQLAlchemy
+from fuzzywuzzy import fuzz
 from manage import Output, Input
 from pytz import timezone
 from sqlalchemy import create_engine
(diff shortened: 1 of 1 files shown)


dmalan:
@brianyu28 want to take a look before merging?

brianyu28:
Hm I think we need to do some rethinking in general as to how best to handle `/review`. The number of uncaught error messages that have been fed through `help50` have gotten pretty numerous in the months since launch, and my browser's been crashing upon trying to access `/review` because there are just too many of them-- @dmalan, do you have the same issue or are you able to scroll through them all?

dmalan:
Hm, yeah, really hangs for me too. Want to whip up some quick pagination? Should be pretty easy by adding an HTTP parameter and using http://getbootstrap.com/components/#pagination? Alternatively, we can just use a command line occasionally. Or just have /review spit out a CSV or TXT file for local review?

brianyu28:
@coltonoscopy Since pull request reminders keep reminding me of this, going to close it for now-- the size of `/review` has gotten so big that this fuzzy matching algorithm is likely going to take too long to be useful in the web interface.

Issue #195, opened 2016-09-25
ls: cannot access asdf: No such file or directory


dmalan:
```
$ ls asdf
ls: cannot access asdf: No such file or directory
```

PR #218, opened 2016-09-29
adds check for ls

coltonoscopy:
Attempts to solve Issue #195 but doesn't work when tested through terminal. Is mirrored closely after already existing tests, and regex was verified in a separate test file


dmalan:
:shipit:

PR #219, opened 2016-09-30
removed matcher that seemed potentially inaccurate

dmalan:


dmalan:
@brianyu28 just noticed this one, which feels now like it's likely to be offering misleading advice. Am inclined to delete, per this PR?

brianyu28:
Hm, not totally opposed to deleting it, but this one happened to show up a lot early on, and always for the same reason: whenever code begins with `int main(void);` with the semicolon at the end, this is the error that's triggered. Feels like a missed opportunity to not try to handle that scenario, though perhaps there's a way to reword so as not to mislead in cases where the error had a different cause?

dmalan:
https://github.com/cs50/help50/pull/20

dmalan:
Ah, makes sense then. Okay, I just simplified the language a bit (to avoid saying "header").

dmalan:
https://github.com/cs50/help50/commit/b038f8695a5306d2a47b751c5bfe40ea72c746cf

PR #220, opened 2016-09-30
Added handling for variable uninitialized when used here

haelannaleah:
Handle the following error:
```
    # $ clang foo.c
    # foo.c:6:20: error: variable 'x' is uninitialized when used here [-Werror,-Wuninitialized]
    #     printf("%d\n", x);
    #                    ^
```

### DIFF
--- a/helpers/clang.py
+++ b/helpers/clang.py
@@ -580,13 +580,16 @@ def help(lines):
     matches = match(r"variable '(.*)' is uninitialized when used here", lines[0])
     if matches:
         response = [
-            "It looks like you're trying to use the variable `{}` on line {} of `{}`.".format(matches.group(3), matches.group(2), matches.group(1)),
-            "However, on that line, the variable `{}` doesn't have a value yet.".format(matches.group(3)),
-            "Be sure to assign a value to `{}` before trying to access its value.".format(matches.group(3))
+            "It looks like you're trying to use the variable `{}` on line {} of `{}` in the".format(matches.group(3), matches.group(2), matches.group(1)),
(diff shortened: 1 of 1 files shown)


helpers/clang.py
```diff
@@ -580,14 +580,17 @@ def help(lines):
     matches = match(r"variable '(.*)' is uninitialized when used here", lines[0])
     if matches:
         response = [
-            "It looks like you're trying to use the variable `{}` on line {} of `{}`.".format(matches.group(3), matches.group(2), matches.group(1)),
-            "However, on that line, the variable `{}` doesn't have a value yet.".format(matches.group(3)),
-            "Be sure to assign a value to `{}` before trying to access its value.".format(matches.group(3))
+            "It looks like you're trying to use the variable `{}` on line {} of `{}` in the".format(matches.group(3), matches.group(2), matches.group(1)),
+            "condition of a loop. However, on that line, the variable `{}` doesn't have a".format(matches.group(3)),
+            "value yet, so we can't use it to check to see if our loop should continue.",
+            "Have you considered using a do-while loop instead?"
```
dmalan: Careful, this matcher seems to be based only on this information:

```
variable '(.*)' is uninitialized when used here
```

Can't draw these conclusions about loops!

haelannaleah: I can't find another instance in which this error occurs. Any other ideas?

haelannaleah: Sorry! Looks like something got messed up on my branch - that's the wrong error! Fixing it now.

PR #231, opened 2016-10-09
Fix out-of-bounds error on valgrind matcher

brianyu28:
Addressing sysadmins reported issue 7946.

The current `valgrind` matcher fails whenever `all heap blocks are freed` but there are other errors, because it's searching for the line that says `0 errors from 0 contexts` and it searches one past the last line. Modified to stop searching after it hits the last line.


dmalan:
Sorry, missed this!

PR #248, opened 2017-06-04
Port help50 to Python

crossroads1112:
Styling is now PEP8 compliant according to the `pep8` tool.

### DIFF
--- a/help50
+++ b/help50
@@ -1,59 +1,105 @@
-#!/bin/bash -i
+#!/usr/bin/env python
+from __future__ import print_function
+import io
+import os
+import signal
+import subprocess
+import sys
+import textwrap
 
-# listen for ctrl-c
-trap "{ echo; exit 1; }" INT
+import requests
+import termcolor
 
-# non-interactive stdin
-# http://stackoverflow.com/a/18766794/5156190
-if [[ $# -eq 0 ]]; then
 
-    # allow input via pipe, disallow interactive input (as potentially confusing)
-    if [[ -t 0 ]]; then
-        echo "Careful, you forgot to tell me with which command you need help!"
-        exit 1
-    else
-        script=$(cat)
-        echo
-    fi
+def main():
+    # Return 1 on Ctrl-C.
+    signal.signal(signal.SIGINT, lambda signum, frame: sys.exit(1))
 
-# interactive stdin
-elif [[ $# -eq 1 && "$1" == "-i" ]]; then
-    shift
-    script=$(cat)
-    echo
+    script = ""
+    print("")
 
-# cmd
-# capture and echo both stdout and stderr, http://stackoverflow.com/a/12451419/5156190
-# preserve colorized output, http://stackoverflow.com/a/1402389
-else
-    exec 3>&1
-    echo
-    script=$(stdbuf -e 0 -o 0 bash -c -i "script --command \"bash -l -c \\\"$*\\\"\" --quiet --return /dev/null" |& tee >(cat - >&3))
-    echo
-fi
+    # Non-interactive stdin
+    if len(sys.argv) == 1:
 
-# URL for POST
-if [[ -z "$APP_URL" ]]; then
-    APP_URL=https://help.cs50.net/
-fi
+        # Allow input via pipe, disallow interactive input (as potentially
+        # confusing)
+        if sys.stdin.isatty():
+            print("Careful, you forgot to tell me"
+                  "with which command you need help!")
+            sys.exit(1)
+        else:
+            script = sys.stdin.read()
 
-# POST /
-echo -e -n "\033[33m";
-echo -e -n "Asking for help..."
-stdout=$(curl \
-    --data-urlencode "cmd=$*" \
-    --data-urlencode "format=ans" \
-    --data-urlencode "script=$script" \
-    --data-urlencode "username=$C9_USER" \
-    --fail \
-    --request POST \
-    "$APP_URL" \
-    2> /dev/null
-)
-status=$?
-printf "\r%${COLUMNS}s\r"
-if [[ $status -ne 0 ]]; then
-    stdout="Ack, there seems to be a bug in help50! Please let sysadmins@cs50.harvard.edu known with which error you need help!"
-fi
-echo -e "$stdout" | fold -s -w "$COLUMNS"
-echo -e "\033[0m"
+    # Interactive stdin
+    elif len(sys.argv) == 2 and sys.argv[1] == "-i":
+        script = sys.stdin.read()
+
+    # Command
+    else:
+        # Capture stdout and stderr from process, and print it out.
+        with io.StringIO() as iostr:
+            tee(" ".join(sys.argv[1:]), iostr, sys.stdout)
+            script = iostr.getvalue()
+
+    termcolor.cprint("\nAsking for help...", "yellow", end="")
+    sys.stdout.flush()
+
+    data = {
+        "cmd": " ".join(sys.argv[1:]),
+        "format": "ans",
+        "script": script,
+        "username": os.environ.get("C9_USER")
+    }
+
+    # Read help50 url and from environment variable,
+    # defaulting to help.cs50.net.
+    app_url = os.environ.get("APP_URL") or "https://help.cs50.net/"
+
+    # Read number of columns in terminal from environment variable, defaulting
+    # to 100.
+    columns = os.environ.get("COLUMNS") or 100
+
+    # Connect to server and print response,
+    # showing error message if unable to.
+    try:
+        # Connect to server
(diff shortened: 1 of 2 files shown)


dmalan:
Differs from current version's output (aka `./old`) in that it has an extra newline at top.

screen shot 2017-06-24 at 2 44 45 pm

dmalan:
ANSI color codes not preserved in new version:

screen shot 2017-06-24 at 2 47 48 pm

dmalan:
Upon receiving Ctl-C, let's output a `\n` so that shell prompt ends up on new line, not like:

dmalan:
Missing space:

screen shot 2017-06-24 at 2 58 35 pm

help50
```diff
@@ -1,59 +1,105 @@
...
+import subprocess
+import sys
+import textwrap
+
+import requests
+import termcolor
+
+
+def main():
+    # Return 1 on Ctrl-C.
```
dmalan: Let's maintain consistent style with other Python apps: blank line above command, uncapitalized `return` and no trailing `.`. Elsewhere too.

help50
```diff
@@ -1,59 +1,105 @@
...
+    signal.signal(signal.SIGINT, lambda signum, frame: sys.exit(1))
+
+    script = ""
+    print("")
+
+    # Non-interactive stdin
```
dmalan: `non-interactive stdin`

help50
```diff
@@ -1,59 +1,105 @@
...
+import requests
+import termcolor
+
+
+def main():
+    # Return 1 on Ctrl-C.
+    signal.signal(signal.SIGINT, lambda signum, frame: sys.exit(1))
+
+    script = ""
+    print("")
```
dmalan: Needed? See comments in PR.

help50
```diff
@@ -1,59 +1,105 @@
...
+
+    # Non-interactive stdin
+    if len(sys.argv) == 1:
+
+        # Allow input via pipe, disallow interactive input (as potentially
+        # confusing)
+        if sys.stdin.isatty():
+            print("Careful, you forgot to tell me"
+                  "with which command you need help!")
```
dmalan: No need to wrap such short lines.

help50
```diff
@@ -1,59 +1,105 @@
...
+            sys.exit(1)
+        else:
+            script = sys.stdin.read()
+
+    # Interactive stdin
+    elif len(sys.argv) == 2 and sys.argv[1] == "-i":
+        script = sys.stdin.read()
```
dmalan: Let's use ArgParser for consistency, even though there's just one switch. E.g., https://github.com/cs50/submit50/blob/develop/submit50.py#L106-L112.

help50
```diff
@@ -1,59 +1,105 @@
...
+
+    # Command
+    else:
+        # Capture stdout and stderr from process, and print it out.
+        with io.StringIO() as iostr:
+            tee(" ".join(sys.argv[1:]), iostr, sys.stdout)
+            script = iostr.getvalue()
```
dmalan: If `help50` is being run on a slow-running process that outputs lots of output, will the output appear on screen immediately or get buffered (until program terminates)? Can we ensure the former happens so students see output in realtime?

help50
```diff
@@ -1,59 +1,105 @@
...
+        "username": os.environ.get("C9_USER")
+    }
+
+    # Read help50 url and from environment variable,
+    # defaulting to help.cs50.net.
+    app_url = os.environ.get("APP_URL") or "https://help.cs50.net/"
+
+    # Read number of columns in terminal from environment variable, defaulting
+    # to 100.
+    columns = os.environ.get("COLUMNS") or 100
```
dmalan: Let's default to 80.

help50
```diff
@@ -1,59 +1,105 @@
...
+        termcolor.cprint(wrap("Ack, there seems to be a bug in help50!"
+                              "Please let sysadmins@cs50.harvard.edu know"
+                              "with which error you need help!",
+                              columns),
+                         "yellow")
+        sys.exit(1)
+    else:
+        # Overwrite previous line with spaces
+        print("\r{}\r".format(" " * columns), end="")
+        termcolor.cprint(wrap(r.text.encode("utf-8").decode("unicode_escape"),
```
dmalan: Is `unicode_escape` necessary here? What problem does it address, given that you've specified `utf-8`?

help50
```diff
@@ -1,59 +1,105 @@
...
+
+
+def wrap(string, columns):
+    """ Wraps a string to the specified number of columns,
+        preserving blank lines.
+    """
+    return "\n".join(
+        ("\n".join(textwrap.wrap(line, columns))
+         for line in string.split("\n")))
+
```
dmalan: Should this more properly be a list comprehension with `[]` and not a tuple?

help50
```diff
@@ -1,59 +1,105 @@
...
+
+def tee(command, *outputs):
+    """Executes command, piping stdout and stderr to *outputs."""
+
```
dmalan: Why blank line?

help50
```diff
@@ -1,59 +1,105 @@
...
+def wrap(string, columns):
+    """ Wraps a string to the specified number of columns,
+        preserving blank lines.
+    """
+    return "\n".join(
+        ("\n".join(textwrap.wrap(line, columns))
+         for line in string.split("\n")))
+
+
+def tee(command, *outputs):
```
dmalan: See question above about realtime.

help50
crossroads1112: Will do. PEP-8 is a bit ambiguous on this front:


> Comments should be complete sentences. If a comment is a phrase or sentence, its first word should be capitalized, unless it is an identifier that begins with a lower case letter (never alter the case of identifiers!).

> If a comment is short, the period at the end can be omitted. Block comments generally consist of one or more paragraphs built out of complete sentences, and each sentence should end in a period.

dmalan: Grrr, okay, let's do:

`This is a short comment`

and

`This is a short comment. But a second sentence too.`

for now.

help50
crossroads1112: This actually isn't a tuple, but rather a generator comprehension, which (should) be more efficient in this case as it doesn't require allocating additional memory. Generator comprehensions are documented [here](https://www.python.org/dev/peps/pep-0289/).

dmalan: Fancy, very nice.

help50
crossroads1112: As it is currently written, it is line buffered, but it does not wait for program termination otherwise. E.g. `help50 find ~` shows the find's output in real time.

help50
crossroads1112: Output looks like this otherwise:

http://i.imgur.com/RBbzUa4.png

I don't fully understand why, but the ANSI codes that dictate the background color do not display without this line.

help50
```diff
@@ -1,59 +1,105 @@
...
+#!/usr/bin/env python
+from __future__ import print_function
+import io
+import os
+import signal
+import subprocess
+import sys
+import textwrap
+
+import requests
```
crossroads1112: From PEP-8
```
Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

Imports should be grouped in the following order:

    standard library imports
    related third party imports
    local application/library specific imports

You should put a blank line between each group of imports.
```

help50
crossroads1112: Have we decided on an appropriate max line length? PEP-8 specifies 79 characters (in which case the unwrapping the lines puts us over)

help50
crossroads1112: And then an even shorter comment (e.g. `non-interactive stdin`) is not capitalized at all?

help50
dmalan: Can we turn off buffering and output anytime `stdout` or `stderr` is available, as via https://github.com/cs50/python-cs50/blob/develop/src/cs50/cs50.py#L6-L22? That way, `help50` can work with interactive programs too that might have prompts.

help50
dmalan: Interesting.

help50
dmalan: 👍

help50
dmalan: Sorry, meant to delete that comment. Checked PEP after hitting Enter. 79 aok.

help50
dmalan: So, based on PEP, seems that should be `Non-interactive stdin` after all, since it's a "phrase"?

dmalan: So, based on PEP, seems that should be `Non-interactive stdin` after all. Though it's indeed a bit awkwardly worded that they introduce "phrase" after saying that comments should be "sentences."

dmalan:
Thank you!

PR #249, opened 2017-06-06
Travis

kzidane:
based on #248! 

cc @crossroads1112

PR #254, opened 2017-06-28
Resolves #252: Remove hardcoded python2 and fix missing executable error

### LINKED ISSUE #252: un-hardcode python 2
https://github.com/cs50/help50/blob/develop/help50#L1

crossroads1112:
Address #249 and #252

### DIFF
--- a/help50
+++ b/help50
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 
 from __future__ import print_function
 import io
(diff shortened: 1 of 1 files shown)


help50
```diff
@@ -40,10 +40,18 @@ def main():
...
-            tee(proc, iobytes, stdout)
-            proc.wait()
-            script = iobytes.getvalue().decode("utf-8")
+            try:
+                proc = ptyprocess.PtyProcess.spawn(sys.argv[1:])
+            except FileNotFoundError:
+                # Executable was not found, replicate shell error message
+                script = "{0}: {1}: command not found".format(
+                            os.path.basename(os.environ.get("SHELL") or "bash"),
+                            sys.argv[1])
```
dmalan: Why not just capture stderr and show that rather than recreate a message?

help50
```diff
@@ -40,10 +40,18 @@ def main():
...
+                print(script)
+            else:
+                tee(proc, iobytes, stdout)
+                proc.wait()
+                script = iobytes.getvalue().decode("utf-8")
```
dmalan: Do add comments for these lines and above so we remember long-term why this is so fancy.

help50
crossroads1112: Because we aren't actually running the command in a shell so there's nothing to actually create that error message. It is a bit hackish, granted, but short of running the command as an argument of `bash -c` which feels unnecessary, I can't think of a better way.

help50
```diff
@@ -40,10 +40,25 @@ def main():
...
+            # executable isn't found
+            try:
+                proc = ptyprocess.PtyProcess.spawn(sys.argv[1:])
+
+            # If the executable isn't found, we need to replicate
+            # the shell's error message
+            except FileNotFoundError:
+                script = "{0}: {1}: command not found".format(
+                    os.path.basename(os.environ.get("SHELL") or "bash"),
+                    sys.argv[1])
```
dmalan: If `-c` will do the trick, let's go that route to capture actual `stderr`? E.g., on Mac OS, the output's a bit different, `-bash: foo: command not found`, and replicating an error message we can just capture platform-independently feels more robust, especially since we pattern-match on `stderr` in `help50`'s back end.

crossroads1112: Do you want it to always run with `bash -c` or only if we encounter a `FileNotFoundError`

dmalan: Any reason not to just run it with `-c` always for simplicity?

PR #272, opened 2019-02-16
run login shell

kzidane:
To source `/etc/profile.d/ide50.sh` and others.


dmalan:
If this works okay in v3, v4, and sandbox, aok to roll out!

PR #281, opened 2019-12-25
Remove Pexpect

### LINKED ISSUE #275: pexpect buggy on Linux
https://github.com/pexpect/pexpect/issues/120

ashenm:
Resolves #275 and constricts Windows support disjunction

### DIFF
--- a/help50/__main__.py
+++ b/help50/__main__.py
@@ -8,11 +8,11 @@ import sys
 import tempfile
 import textwrap
 import traceback
+import subprocess
 
 from argparse import ArgumentParser, REMAINDER
 import lib50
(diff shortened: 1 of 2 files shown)


kzidane:
The reason why pexpect is used is to trick the process into thinking that it's running in a terminal to preserve things like colored output. Although using `subprocess.run` or `subprocess.check_output` would work, it wouldn't preserve colored output. I tried removing pexpect per https://github.com/cs50/help50/pull/282 and using `check_output` on Windows instead, so that version of `help50` should work on Windows now albeit colors outputted by the passed command probably won't work.

PR #282, opened 2020-01-08
removed pexpect, fixed windows

kzidane:


kzidane:
Looks like it's different as you guessed.

crossroads1112:
@kzidane, I thought this was a pull request into develop. Why did you merge directly into master instead of going through develop? Now develop is in an inconsistent state.

kzidane:
My bad. I thought develop had unfinished work at the time. Are we all set?

crossroads1112:
Nbd, yeah I think we should be.

PR #290, opened 2021-08-08
update lib50 requirement, bump version

rongxin-liu:
Would the latest version of lib50 touch help50? I tested help50 with the latest version of lib50, and it seemed working fine, but I just want to double-check in case I overlooked something.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ setup(
     ],
     license="GPLv3",
     description="This is help50, a command-line tool that helps students understand error messages.",
-    install_requires=["colorama", "termcolor", "lib50>2,<3"],
+    install_requires=["colorama", "termcolor", "lib50>=3,<4"],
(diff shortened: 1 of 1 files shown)


rongxin-liu:
PS: cs50/cli was using help50 version 3.0.2 instead of the latest version (3.0.4) due to lib50 requirement.

dmalan:
@Jelleas @cmlsharp any concerns here?

Jelleas:
Nope, should be fine! The only thing the latest lib50 changed was an option to set a custom honesty prompt. As far as I know, it's just submit50 that needed changes.

rongxin-liu:
> Nope, should be fine! The only thing the latest lib50 changed was an option to set a custom honesty prompt. As far as I know, it's just submit50 that needed changes.

Sounds good, thank you!

